PR #1, opened 2016-09-06
S2I Play Builder

sabre1041:
What does this PR do?

Adds an S2I [Play Framework](https://www.playframework.com) builder

How should this be manually tested?

Execute the following in the `s2i-play` directory

```
oc new-build registry.access.redhat.com/rhel7.2 --name=s2i-play --binary=true
oc start-build s2i-play --from-dir=.
```

Once complete, an example app can be used to validate the functionality

```
oc new-app s2i-play~https://github.com/playframework/play-ebean-example.git
```

An additional environment needs to be added to the deploymentconfig

```
oc env dc/play-app JAVA_OPTS_EXT=-Dplay.evolutions.db.default.autoApply=true
```

Who would you like to review this?

/cc @etsauer


etsauer:
@sabre1041 this is where we need to discuss what goes into the README...

sabre1041:
@etsauer I have updated the README to include a full demo writeup. Please review

etsauer:
@sabre1041 I think it might be worth building out a quickstart template that someone could import into OpenShift to provide Play 'as a service'. What do you think?

otherwise this looks really good. would like others to review this as well.

sabre1041:
@etsauer really like that idea. I'll create a template that sets up the builder build and and a sample application. once the builder build completes, it will trigger the app build. a nice continuous delivery implementation

sabre1041:
@etsauer added rhel based template

quickstart-templates/rhel-templats.json
```diff
@@ -0,0 +1,348 @@
...
+					"kind": "ImageStream",
+					"apiVersion": "v1",
+					"metadata": {
+						"name": "rhel7",
+						"labels": {
+							"build": "${PLAY_BUILDER_NAME}"
+						}
+					},
+					"spec": {
+						"dockerImageRepository": "registry.access.redhat.com/rhel7/rhel"
```
etsauer: I believe the standard now is to leave the registry host out of the image URL, and let the local docker config work it out between registry.access... and docker.io

quickstart-templates/rhel-templats.json
```diff
@@ -0,0 +1,348 @@
...
+					"description": "Path within Git project to build containing the Play Framework application; empty for root project directory.",
+					"displayName": "Play application context directory.",
+					"name": "PLAY_APPLICATION_CONTEXT_DIR",
+					"value": "",
+					"required": false
+				},
+				{
+					"name": "PLAY_APPLICATION_HOSTNAME",
+					"displayName": "Play application route hostname.",
+					"description": "Custom hostname for Play Framework service route.  Leave blank for default hostname, e.g.: \u003capplication-name\u003e-\u003cproject\u003e.\u003cdefault-domain-suffix\u003e"
```
etsauer: description here mentions a default value, but no value is defined.

s2i-play/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM rhel7.2
+
+MAINTAINER Andrew Block <ablock@redhat.com>
+
+LABEL io.k8s.description="Platform for building Play Framework Applications" \
+    io.k8s.display-name="Play Framework" \
+    io.openshift.expose-services="8080:http" \
+    io.openshift.tags="builder,webserver,html,play" \
+    io.openshift.s2i.scripts-url="image:///usr/libexec/s2i" \
+    io.openshift.s2i.destination="/opt/app-root"
```
etsauer: I know origin stuff doesn't do this, but should we include a label with our source repo?

quickstart-templates/rhel-templats.json
sabre1041: This is the default action for a route. If a hostname is not specified, it will use the default OpenShift pattern

etsauer: ok, disregard then.

quickstart-templates/rhel-templats.json
sabre1041: That only applies to Dockerfiles. ImageStreams still need full address since the API performs the resolution and not the docker daemon

quickstart-templates/rhel-templats.json
```diff
@@ -0,0 +1,348 @@
+{
```
sabre1041: Corrected

s2i-play/Dockerfile
sabre1041: If these images are being pushed to DockerHub, they will most likely have the source code associated so there is less of a need

sabre1041:
@etsauer added centos templates and made some adjustments based on prior comments

s2i-play/README.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+A result similar to the following should appear:
+
+```
+NAME               READY     STATUS             RESTARTS   AGE
+play-app-1-<random>   0/1       CrashLoopBackOff   6          8m
+```
+
+The *CrashLoopBackOff* status indicates that there is an error starting the pod. Attempting to view the pod logs by executing `oc logs play-app-1-<random> provides insight into the pod failures. This provides an opportunity for users to have first hand experiencing troubleshooting OpenShift applications.
```
etsauer: no closing back tic here.

etsauer:
@sabre1041 haven't quite figured out this new review process... but one more small syntax issue above, then it's good to go.

sabre1041:
@etsauer should be good to go

PR #8, opened 2016-11-11
Updated s2i-java README to include full BOM

sabre1041:
Added a BOM for s2i-java image to resolve #8 

CC: @etsauer @raffaelespazzoli


raffaelespazzoli:
the documentation looks good to me.
one question on the template. it seems to b designed to build a play framework app. Why not let it be a generic template for any maven app? Probably I'm missing something here, sorry if it is a dumb question.

sabre1041:
The template file is actually a list of multiple templates. This list includes both a template for play and java, among others

etsauer:
lgtm. merging

PR #26, opened 2017-02-25
Added deprecated notice to s2i-java image

sabre1041:
#### What is this PR About?
Adds deprecation notice to s2i-java image as supported image is now available

#### How do we test this?
N/A


sabre1041:
bump @redhat-cop/containerize-it @etsauer @JaredBurck

PR #47, opened 2017-10-17
Deprecate java and httpd examples

sabre1041:
#### What is this PR About?
Deprecated java and httpd examples as official Red Hat images have been published


oybed:
@sabre1041 @etsauer This bit our Labs demo as we were relying on these repos. When we say “deprecate”, we need to make sure we don’t just remove the content, but truly deprecate.

Issue #56, opened 2017-11-29
openshift-applier support for provisioning


sabre1041:
Support for utilizing the [openshift-applier](https://github.com/redhat-cop/casl-ansible/tree/master/roles/openshift-applier) to provision applications

* Creation of inventory for quickstarts
* Documentation on usage
* Continuation of support without openshift-applier

pabrahamsson:
This effort has been completed

PR #63, opened 2018-01-20
Use openshift-applier to build and deploy s2i-gows

pabrahamsson:
#### What is this PR About?
Use openshift-applier to build and deploy s2i-gows

#### How do we test this?
See README in PR

#### Is there a relevant Trello card or Github issue open for this?
This will help close out [issue #56](https://github.com/redhat-cop/containers-quickstarts/issues/56)
https://trello.com/c/RFskWdVg/20-add-openshift-applier-support-to-container-quickstart-examples


s2i-gows/files/builds/gows.yml
```diff
@@ -0,0 +1,151 @@
+apiVersion: v1
```
sabre1041: Should this be a template instead of a list of object?

pabrahamsson: Thank you @sabre1041 
I was deliberating whether it should be a template or not. The first iteration was a template but with so many things that are custom I decided to hardcode things and make it a list.
Let me take another stab at a template as it would probably be the right thing:tm: to do (at least make the site you want to deploy a parameter, `demo` might not cut it for everyone ;)

pabrahamsson:
Updated builds to use template from conversation with @sabre1041

s2i-gows/README.md
```diff
@@ -4,12 +4,15 @@
 
 ## Quickstart
 
-Here's a simple on-liner to get you started.
+1. Clone this repository:
+   `git clone https://github.com/redhat-cop/containers-quickstarts`
+2. Run `ansible-galaxy install -r requirements.yml --roles-path=roles`
+3. `cd containers-quickstarts/s2i-gows`
```
sabre1041: 2 and 3 need to be swapped.

pabrahamsson: haha, yeah that would help, updated.
Thank you @sabre1041

pabrahamsson:
Hey @sabre1041 do you think we're good to merge here? I believe I've addressed the last comment you had.

sabre1041:
LGTM. Thanks!

PR #73, opened 2018-02-19
Use openshift-applier to build and deploy Liberty

pabrahamsson:
#### What is this PR About?
Use openshift-applier to build and deploy an s2i image to Liberty

#### How do we test this?
See README in PR

#### Is there a relevant Trello card or Github issue open for this?
This will help close out [issue #56](https://github.com/redhat-cop/containers-quickstarts/issues/56) and [Add Openshift-applier to container-quickstart examples](https://trello.com/c/RFskWdVg/20-add-openshift-applier-support-to-container-quickstart-examples)

### DIFF
--- a/s2i-liberty/applier/templates/build.yml
+++ b/s2i-liberty/applier/templates/build.yml
@@ -1,0 +1,187 @@
+apiVersion: v1
+kind: Template
+labels:
+  template: liberty-builder
+metadata:
+  name: liberty
+objects:
+- apiVersion: v1
+  kind: ImageStream
+  metadata:
+    labels:
+      build: s2i-liberty
+    name: s2i-liberty
+  spec:
+    lookupPolicy:
+      local: false
+- apiVersion: v1
+  kind: ImageStream
+  metadata:
+    labels:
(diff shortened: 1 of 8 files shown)


s2i-liberty/applier/templates/build.yml
```diff
@@ -0,0 +1,187 @@
...
+      local: false
+    tags:
+    - annotations:
+        openshift.io/imported-from: websphere-liberty:webProfile7
+      from:
+        kind: DockerImage
+        name: websphere-liberty:webProfile7
+      importPolicy: {}
+      name: webProfile7
+- apiVersion: v1
```
etsauer: why are we doing this vs just referencing imagestream/jboss-eap70-openshift in the openshift project?

pabrahamsson: I'd say the real reason is probably because I was being lazy and only converted what was already in the existing template. I'll take a closer look or if @raffaelespazzoli might have some input since he did the original template.

pabrahamsson:
Updated per your comments @etsauer

Issue #78, opened 2018-03-07
Migrate Jenkins Slave from Labs CI/CD


sherl0cks:
Also remove ansible stacks slave as that is out of date.

Per discussion with @sabre1041 @etsauer

springdo:
Happy to take on this work guys. I'll fork the repo and start moving my outstanding PR from labs-ci-cd for merging here.

pabrahamsson:
Fixed by #80

PR #80, opened 2018-03-13
Feature/jenkins slave golang

springdo:
#### What is this PR About?
#78 
Adding a new jenkins slave to be used in OpenShift for running Golang Build & Tests

@sherl0cks @etsauer @sabre1041 - what do you guys think of this? If you're cool with this format; I'll start moving all the other `jenkins-slave-*` from labs-ci-cd. Would like to see this one approved before I begin that process as each of the slaves follows the same pattern.

#### How do we test this?
Run the command to apply in the README.md

cc @redhat-cop/containerize-it

### DIFF
--- a/jenkins-slaves/jenkins-slave-golang/Dockerfile
+++ b/jenkins-slaves/jenkins-slave-golang/Dockerfile
@@ -1,0 +1,21 @@
+FROM openshift/jenkins-slave-base-centos7:latest
+# FROM openshift/jenkins-slave-base-rhel7:latest
+
+USER root
+ADD https://dl.google.com/go/go1.9.3.linux-amd64.tar.gz /usr/
+RUN mkdir -p /usr/src/go/src/redhat && \
+    tar -xzf /usr/go1.9.3.linux-amd64.tar.gz && \
+    mv $(pwd)/go /usr/local/ && \
+    chown -R 1001 /usr/src/go && \
+    chown -R 1001 /usr/local/go
+
+USER 1001
+
+ENV GOROOT /usr/local/go
+ENV GOPATH /usr/src/go
+
+ENV PATH $GOPATH/bin:$GOROOT/bin:$PATH
+
+# useful for verification of install
+# RUN go version
+# RUN go env
--- a/jenkins-slaves/jenkins-slave-golang/README.md
+++ b/jenkins-slaves/jenkins-slave-golang/README.md
@@ -17,7 +17,7 @@ oc process -f ../templates/jenkins-slave-generic-template.yml \
 For all params see the list in the `../templates/jenkins-slave-generic-template.yml` or run `oc process --parameters -f ../templates/jenkins-slave-generic-template.yml`.
 
 ## Jenkins
-Add a new Kubernetes Container template called `jenkins-slave-golang` and specify this as the node when running builds. There are path issues with Jenkins permissions and Go when trying to run a build so easiest way to fix this is to setup the GOLANG path to be same as the WORKSPACE
+Add a new Kubernetes Container template called `jenkins-slave-golang` (if you've build and pushed the container image locally) and specify this as the node when running builds. If you're using the template attached; the `role: jenkins-slave` is attached and Jenkins should automatically discover the slave for you. Further instructions can be found [here](https://docs.openshift.com/container-platform/3.7/using_images/other_images/jenkins.html#using-the-jenkins-kubernetes-plug-in-to-run-jobs). There are path issues with Jenkins permissions and Go when trying to run a build so easiest way to fix this is to setup the GOLANG path to be same as the WORKSPACE
 ```
 export GOPATH=${WORKSPACE}
 go get -v -t ./...
--- a/jenkins-slaves/jenkins-slave-golang/README.md
+++ b/jenkins-slaves/jenkins-slave-golang/README.md
@@ -1,0 +1,27 @@
+# jenkins-slave-golang
+Provides a docker image of the golang runtime for use as a Jenkins slave.
+
+## Build local
+`docker build -t jenkins-slave-golang .`
+
+## Run local
+For local running and experimentation run `docker run -i -t jenkins-slave-golang /bin/bash` and have a play once inside the container.
+
+## Build in OpenShift
+```bash
+oc process -f ../templates/jenkins-slave-generic-template.yml \
+    -p NAME=jenkins-slave-golang \
+    -p SOURCE_CONTEXT_DIR=jenkins-slaves/jenkins-slave-golang \
+    | oc create -f -
+```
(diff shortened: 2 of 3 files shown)


jenkins-slaves/jenkins-slave-golang/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM openshift/jenkins-slave-base-centos7:latest
+# FROM openshift/jenkins-slave-base-rhel7:latest
```
sherl0cks: https://github.com/redhat-cop/containers-quickstarts/pull/80/files#diff-b06b46547848dce57f3d821c78624575R21 will run this build with a RHEL base in OCP. There can be times where there is a material difference between centos and rhel, so the Dockerfile will work in one and not the other. I don't see anything here like that (e.g. EPEL is the biggie), but its good to know

sherl0cks:
@etsauer @pabrahamsson @sabre1041 how do you want review / maintenance of these slaves to work? I'm happy to review and merge if that useful, though Im biased a bit as some of the migrated work is mine. Also, how do want to handle CI?

jenkins-slaves/jenkins-slave-golang/Dockerfile
springdo: @sherl0cks  - yeah agreed the `FROM` is overwritten in OCP land when running a build and Im happy for that, aware of EPEL and such. I do find it useful to be able to build locally in Centos all the same as I don't have to mess around with access to rhel repos.

springdo: @pabrahamsson  -  I could change it to to `FROM registry.access.redhat.com/<distro>`; that's no problem. This `FROM` block is only used for local building / testing, so wanted to keep it `centos` based. When you run the template in OCP land it gets overwritten by the RHEL Image as per @sherl0cks  comment

jenkins-slaves/jenkins-slave-golang/README.md
```diff
@@ -0,0 +1,27 @@
...
+```bash
+oc process -f ../templates/jenkins-slave-generic-template.yml \
+    -p NAME=jenkins-slave-golang \
+    -p SOURCE_CONTEXT_DIR=jenkins-slaves/jenkins-slave-golang \
+    | oc create -f -
+```
+For all params see the list in the `../templates/jenkins-slave-generic-template.yml` or run `oc process --parameters -f ../templates/jenkins-slave-generic-template.yml`.
+
+## Jenkins
+Add a new Kubernetes Container template called `jenkins-slave-golang` and specify this as the node when running builds. There are path issues with Jenkins permissions and Go when trying to run a build so easiest way to fix this is to setup the GOLANG path to be same as the WORKSPACE
```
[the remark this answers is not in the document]
springdo: Yeah I've seen that before, the auto discovery thing. But have had times in the past where Jenkins has not made the slaves available so included it here. How about I add the role thing and a link to the docs too?

springdo:
I've updated the readme and changed to RHEL in `BUILDER_IMAGE_NAME` in the template.

jenkins-slaves/jenkins-slave-golang/Dockerfile
etsauer: I think that the proper pattern for `FROM` lines should be NOT to specifiy the registry. Allow the build environment to include approved registries in search paths. This allows for enterprise registries to be used as mirrors without code changes.

etsauer:
@springdo @sherl0cks A few thoughts on this...

Firstly, I would prefer we have separate Dockerfiles as the desire is to put these under autobuilds so that users don't have to build the images to consume them. We already to a decent about of auto publishing the CentOS versions of the images to docker cloud, and i don't want to have to make a repo change to set up a build in OpenShift.

As far as CI goes, I think moving forward I would like to move in the direction of including an applier inventory that can both build the image and run a sample pipeline that proves the image is functional.

jenkins-slaves/jenkins-slave-golang/Dockerfile
springdo: Cool - so leave it as it is then @etsauer ? Can we get this puppy in so I can bring across the rest of the slaves I have outstanding on the `labs-ci-cd` repo then?

jenkins-slaves/templates/jenkins-slave-generic-template.yml
```diff
@@ -0,0 +1,94 @@
...
+  spec:
+    tags:
+    - from:
+        kind: DockerImage
+        name: openshift3/jenkins-slave-base-rhel7:${SLAVE_IMAGE_TAG}
+      generation: 2
+      importPolicy: {}
+      name: latest
+      referencePolicy:
+        type: Source
```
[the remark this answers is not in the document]
sabre1041: @pabrahamsson This is the ImageStream for the slave that is being built. It also has the key label `role: jenkins-slave` so that it will automatically be added as a slave in Jenkins

jenkins-slaves/jenkins-slave-golang/Dockerfile
pabrahamsson: I'm cool to leave it as is based on the comment from @etsauer 
For the sake of brevity, what controls the image for the slave build is https://github.com/redhat-cop/containers-quickstarts/pull/80/files#diff-b06b46547848dce57f3d821c78624575R60 rather than what @sherl0cks was referencing.
Unless I've gotten it all backwards we can remove the `jenkins-slave-base-rhel7` ImageStream (which is what was originally referenced).

springdo:
yo @pabrahamsson  - I've removed the image stream for the jenkins base slave as per your request. Sorry took too long; I got carried away with some other stuff.

jenkins-slaves/templates/jenkins-slave-generic-template.yml
```diff
@@ -0,0 +1,78 @@
...
+    labels:
+      build: "${NAME}"
+      type: image
+    name: "${NAME}"
+  spec:
+    nodeSelector:
+    output:
+      to:
+        kind: ImageStreamTag
+        name: "${NAME}:latest"
```
pabrahamsson: I think this should be `name: "${NAME}:${SLAVE_IMAGE_TAG}"`
Otherwise can you remove the `SLAVE_IMAGE_TAG` parameter from the template?

springdo:
hi @pabrahamsson - yeah I was thinking about this; I've not got anything actionable as such part of this PR. We spoke on the call the other day about the idea of CI/ CD for the components in here and other COP libraries but the general consensus was that we needed to figure out what's of use; and CI/CD from there really. As for the splitting of Dockerfile and autobuild; this would be good but at the moment we don't do this... so would be good to just get some stuff into this repo and learn from there I think. 

Thoughts?

pabrahamsson:
@springdo agreed on the let's get started and iterate on this as needed.
@etsauer are you good with merging or did you want any changes around the Dockerfiles?

pabrahamsson:
Merging PR and tracking separate Dockerfiles as #87

PR #93, opened 2018-04-24
Gitlab custom sa

makentenza:
#### What is this PR About?
Add `namespace` parameter so the SA added to `anyuid` scc is created dynamically according to the namespace name.

#### How do we test this?
Follow instructions from README.


gitlab-ce/inventory/group_vars/seed-hosts.yml
```diff
@@ -3,7 +3,9 @@ openshift_cluster_content:
 - object: projectrequest
   content:
   - name: gitlab-project
-    file: "{{ inventory_dir }}/../files/projects/projects.yml"
+    template: "https://raw.githubusercontent.com/redhat-cop/cluster-lifecycle/master/files/projectrequest/template.yml"
```
oybed: @makentenza let's use a specific release here to avoid potential breakage in the future if/when this template is moved/changed.

makentenza: @oybed we haven't created cut releases on `cluster-lifecycle` so far. Can you create one?

oybed: @makentenza hmm, interesting - I was positive we already had one. Anyway, I just cut a `v3.9.0` release:
https://github.com/redhat-cop/cluster-lifecycle/releases/tag/v3.9.0

makentenza: @oybed this is now done

gitlab-ce/files/templates/gitlab-ssl.yml
```diff
@@ -67,7 +67,8 @@ objects:
             command:
             - bash
             - -c
-            - /opt/gitlab/embedded/bin/openssl req -newkey rsa:2048 -nodes -keyout /etc/gitlab/${APPLICATION_HOSTNAME}.key -x509 -days 365 -out /etc/gitlab/${APPLICATION_HOSTNAME}.crt -subj "/C=GB/ST=London/L=London/O=Container and PaaS CoP/CN=${APPLICATION_HOSTNAME}"
+            - test -e "/etc/gitlab/${APPLICATION_HOSTNAME}.crt" && { echo "Custom
```
sabre1041: The space here is causing issue with the template

makentenza: Solved, thanks for reviewing @sabre1041

PR #108, opened 2018-06-12
Add redirection for insecure HTTP access

InfoSec812:
#### What is this PR About?
Add HTTP->HTTPS redirection to GitLab template

#### How do we test this?
Deploy the template and then attempt to access GitLab over HTTP. You should be redirected to the same URL with HTTPS instead.


oybed:
This was already implemented by PR #93

PR #113, opened 2018-06-22
Added gradle s2i builder images

### LINKED ISSUE #41: Need a new s2i builder image that includes gradle
Have a client that has requested a builder image that would include rhel7+/openjdk8+/gradle4+.  They use gradle for their builds and are looking to transition a bunch of new projects to OpenShift, but need this new image.

procrypt:
#### What is this PR About?
RHEL based s2i builder image with maven and gradle.
Fixes #41 
#

#### How do we test this?
`cd s2i-mvn-gradle/`
`make test`

### DIFF
--- a/s2i-mvn-gradle/Dockerfile
+++ b/s2i-mvn-gradle/Dockerfile
@@ -1,0 +1,60 @@
+# openjdk-gradle
+FROM registry.access.redhat.com/rhel 
+
+# TODO: Put the maintainer name in the image metadata
+LABEL maintainer="Abhishek Singh <abhishek@linux.com>" \
+      io.k8s.description="Builder Image for building Java applications with Maven 3.5.3 or Gradle 4.8 on RHEL 7" \
+      io.k8s.display-name="S2I builder 1.0" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,maven-3.5.3,gradle-4.8,java,openjdk-gradle"
+
+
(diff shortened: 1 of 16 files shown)


oybed:
@sabre1041 can you please assist reviewing this PR as I believe you've done something very similar in the past.

s2i-mvn-gradle/Dockerfile
```diff
@@ -0,0 +1,60 @@
+# openjdk-gradle
+FROM registry.access.redhat.com/rhel 
+
+# TODO: Put the maintainer name in the image metadata
+LABEL maintainer="Abhishek Singh <abhishek@linux.com>" \
+      io.k8s.description="Builder Image for building Java applications with Maven 3.5.3 or Gradle 4.8 on RHEL 7" \
+      io.k8s.display-name="S2I builder 1.0" \
```
sabre1041: Can the display name be more descriptive of the gradle builder?

procrypt:
@sabre1041 Thanks for the review. I'll update the PR.

s2i-mvn-gradle/Dockerfile
lpsantil: @sabre1041 One thing that was suggested to me when I had a Spring Boot+Gradle customer was to leverage the current openjdk-s2i base image (to keep full support; removal of bits can degrade support level) and use Maven to trigger the Gradle build.  Yeah, kinda ridiculous but I understood the point about support.

PR #115, opened 2018-07-17
Updated Dockerfile

huddlesj:
#### What is this PR About?
Take care of cases when execute bit is lost when cloning. 

#### How do we test this?
Tested with: oc new-app fedora:latest~https://github.com/huddlesj/containers-quickstarts.git --context-dir=tool-box/ --name=test --strategy=docker


syvanen:
Tested this PR also on my own setup. Makes sense to make sure the execute bit is ensured in DockerFile. For the cases when it's not set on Git.

PR #122, opened 2018-08-06
Update to atomic as base image

dmhumph:
#### What is this PR About?
Move from RHEL to Atomic for base image to create lighter weight images
#### How do we test this?
The repo might need to be changed in the build.yml for the s2i-play project to a current repo with the necessary changes:
https://github.com/dmhumph/containers-quickstarts.git

### DIFF
--- a/s2i-play/applier/templates/build.yml
+++ b/s2i-play/applier/templates/build.yml
@@ -95,7 +95,7 @@ parameters:
   name: PLAY_BUILDER_SOURCE_REPOSITORY_URL
   required: true
   value: https://github.com/dmhumph/containers-quickstarts.git
-  ref: "atomic"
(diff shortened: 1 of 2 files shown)


s2i-play/applier/templates/build.yml
```diff
@@ -99,7 +99,7 @@ parameters:
   displayName: Play builder source branch/tag reference.
   name: PLAY_BUILDER_SOURCE_REPOSITORY_REF
   required: false
-  value: master
+  value: atomic
```
pabrahamsson: Please change back to `master` as this PR will be merged into the master branch of this repo.

pabrahamsson:
Just out of curiosity what difference do you see in image size by this change?

PR #126, opened 2018-08-31
Add Python 3.6 Jenkins slaves

jacobsee:
#### What is this PR About?
Adding Dockerfiles for Python 3.6 Jenkins slaves!

#### How do we test this?
The CentOS Dockerfile can be built anywhere. The RHEL7 one only works on a RHEL host due to package manager subscription requirements. Once built, the images have `python` and `pip` set to the Python 3.6-based installations. `python --version` and `pip --version` can confirm this.


sherl0cks:
@jacobsee looks good to me. did the python repo stuff come from the SCL repo? if so, can you link that for reference.

sherl0cks:
@jacobsee thanks, lgtm. I'll just wait for CI to finish and then merge.

jacobsee:
@sherl0cks Thanks!

PR #135, opened 2018-10-24
Deployment of Hygieia to OpenShift

etsauer:
#### What is this PR About?

#### How do we test this?


hygieia/.openshift/templates/docker-build-git.yml
```diff
@@ -0,0 +1,76 @@
+apiVersion: v1
```
pcarney8: is this the same as the new `build-docker-generic` template?

hygieia/.openshift/templates/projects.yml
```diff
@@ -0,0 +1,17 @@
+---
```
pcarney8: Can we delete this and point to: https://raw.githubusercontent.com/redhat-cop/cluster-lifecycle/v3.9.0/files/projectrequest/template.yml instead?

need to add `NAMESPACE_DISPLAY_NAME: "{{ namespace }}"` to params if using it.

hygieia/.openshift/templates/docker-build-git.yml
etsauer: probably, but since we currently don't have that in a release, I'd rather no point to it yet.

PR #142, opened 2018-11-06
Update the Gradle image from one with JRE to image with JDK

syvanen:
#### What is this PR About?
Updating the Gradle slave to use Jenkins Maven Slave image as base to get the JDK instead of JRE

### DIFF
--- a/jenkins-masters/hygieia-plugin/.openshift/templates/jenkins-s2i.yml
+++ b/jenkins-masters/hygieia-plugin/.openshift/templates/jenkins-s2i.yml
@@ -48,7 +48,7 @@ objects:
           kind: ImageStreamTag
           name: hygieia-plugin:latest
         paths:
-        - destinationDir: "plugins/."
+        - destinationDir: "jenkins-masters/hygieia-plugin/plugins"
           sourcePath: /tmp/src/target/hygieia-publisher.hpi
       type: Image
     strategy:


pabrahamsson:
@syvanen ping

syvanen:
@pabrahamsson Looks like only the upstream version has Gradle not the official image: https://github.com/openshift/jenkins/blob/master/agent-maven-3.5/Dockerfile.rhel7 and https://access.redhat.com/containers/?tab=docker-file#/registry.access.redhat.com/openshift3/jenkins-agent-maven-35-rhel7/images/v3.11.43-2

sabre1041:
@syvanen A couple of things

1. Now that we are not using the Jenkins slave base, when building in OpenShift, we need to specify the Maven slave as the base as a template parameter. (Add `-p BUILDER_IMAGE_NAME= openshift3/jenkins-agent-maven-35-rhel7:v3.11`)
2. Since the ImageStream associated with the builder has the label `role=jenkins-slave`, it is automatically added to Jenkins. However, since  the resulting tag (`v3.11`) is built, it is not properly configured by the Jenkins Sync plugin (only latest). This functionality should exist to pick up the proper ImageStreamTag, however, it appears logic is being handled incorrectly in the plugin. Added an [issue](https://github.com/openshift/jenkins-sync-plugin/issues/278) in the sync plugin project

PR #143, opened 2018-11-10
User RHEL Jenkins Slave Base

InfoSec812:
#### What is this PR About?
Update ZAP slave to be based on RHEL7 Jenkins Slave Base

#### How do we test this?
N/A


InfoSec812:
NVM, this is not right.

PR #147, opened 2018-12-03
Tie Jenkins slave images to match OCP release

makentenza:
#### What is this PR About?
This PR will cover the Jenkins slave images to match the OCP release and solve some existing issues on resultant slaves like npm and others, where we have seen unexpected results when running 'latest' tag of the base images when a new major version is available and the 'latest' tag is moved in the Red Hat Container Catalog.

#### How do we test this?
Build Jenkins slave images using instructions from the repository


makentenza:
Removing WIP after testing all the slave builds in a 3.11 Cluster with RHEL 7.6

makentenza:
@springdo @syvanen this will cover the issue we experienced at VW when using 'latest' tag for Slave images on Clusters not running latest major version. We should create cut releases on this per OCP major versions

syvanen:
@pabrahamsson and @etsauer Any opinions on how we should manage the tagging when it comes to ocp container tags and tagging of this repo?

PR #159, opened 2018-12-19
Add hoverfly command

tylerauerbeck:
#### What is this PR About?
Adds command to hoverfly template that allows it to listen on all interfaces

#### How do we test this?
Spin up hoverfly and verify that you can connect to it


makentenza:
@tylerauerbeck just tested and worked fine now. Can we add another route to the template for the 'admin' interface on port '8888' please? Is an optional thing but I think it'll be helpful for anyone using this, otherwise LGTM

tylerauerbeck:
@makentenza Added a route to the admin interface. Let me know if you see any issues.

pabrahamsson:
@sabre1041 the [docs link](https://docs.hoverfly.io/en/latest/pages/reference/api/api.html) is working now.

PR #168, opened 2019-01-22
Fixing the build template for the Hygieia enabled jenkins

etsauer:
#### What is this PR About?
Templates were broken

#### How do we test this?
jenkins-masters/hygieia-plugin/README.md


sabre1041:
LGTM

PR #174, opened 2019-01-23
Adding build-monitor-plugin to custom jenkins master build

etsauer:
#### What is this PR About?
Apparently the 3.11 jenkins is missing the build-monitor-plugin, which is needed fro the groovy script that sets the root URL. I'm adding it back in.

#### How do we test this?
Spin up the hygieia jenkins quickstart, stand up a pipeline that uses ${JENKINS_URL} and make sure it works.


sabre1041:
LGTM

PR #200, opened 2019-04-01
Owasp Zap Image

haithamshahin333:
#### What is this PR About?
Adds an updated image based on the public owasp zap image to modify permissions to run in openshift. The image automatically starts the zap process which can be used as a proxy when running e2e testing.

Utilizes some work done by @InfoSec812 to have the permissions issues resolved.

#### How do we test this?
readme includes instructions for testing

@sherl0cks @InfoSec812


InfoSec812:
Overall, no serious issues. One concern is that this is based on a Docker Hub image which is not vetted by Red Hat and may cause issues when customers reach out for support. The existing ZAP Dockerfile uses the RHEL Java base image for that reason.

haithamshahin333:
Thanks for the review @InfoSec812 . One point to make here is that unlike the current image in the repo, this is not a jenkins agent - it is just running the zap process.

If the concern is shared by others around the base image (@sherl0cks @pcarney8 @cognizen) I can do some rework here.

cognizen:
> One point to make here is that unlike the current image in the repo, this is not a jenkins agent - it is just running the zap process.

What are the relative merits and drawbacks of running it as a process instead of as an agent?

haithamshahin333:
@cognizen good question. So I see the benefits of running as an agent being that the deployment of zap itself is easier (i.e. just use the usual jenkins agent syntax and now you have a unique instance of zap running, while with this model you need to use the openshift plugin and get smart about the name of the app to ensure it's uniqueness).

However, the downside is that we still need the npm agent running so we can execute the e2e tests. This was why @InfoSec812 was working towards having the zap agent run as a sidecar to the npm agent (example [here](https://github.com/CS-C-BDD-TDD/human-review-ui-tests/blob/master/Jenkinsfile)). I think my thought was that we could avoid that complexity especially since we just want zap to run so we can proxy our tests through it and publish the report it generates.

@InfoSec812 any considerations you have as to why it would be better to run zap as a jenkins agent instead of the model described here?

InfoSec812:
@cognizen @haithamshahin333 Running the ZAP process inside of an existing Jenkins agent (e.g. NodeJS) requires that we install and configure very complex software in each of those build slaves that we want to run ZAP tests... For example, for users of Selenium+JUnit, their Maven agent would need ZAP. For users of NodeJS, they would need to install ZAP in the NodeJS agent. So on and so forth. That's a LOT of extra stuff loaded into each of those agent. Using a separate agent for ZAP means that we just need to set up a podtemplate section in the top of the Jenkinsfile as follows:

```
pipeline {
  agent {
kubernetes {
      label "ruby-go-sidecars-${env.BUILD_ID}"
      cloud "openshift"
      defaultContainer 'jnlp'
      yaml """
apiVersion: v1
kind: Pod
metadata:
  labels:
    pod-template: jenkins-slave-sidecars
spec:
  serviceAccount: jenkins
  serviceAccountName: jenkins
  containers:
  - name: jnlp
    image: docker-registry.default.svc:5000/${currentNamespace}/jenkins-slave-nodejs
    tty: true
  - name: jenkins-slave-zap
    image: docker-registry.default.svc:5000/${currentNamespace}/jenkins-slave-zap
    tty: true
    command:
    - zap.sh
    - -daemon 
    - -host 
    - 0.0.0.0 
    - -port 
    - 8080 
    - -config 
    - api.disablekey=true 
    - -config 
    - api.addrs.addr.regex=true 
    - -config 
    - api.addrs.addr.name=.*
          """
    }
  }
}
```

InfoSec812:
Actually, based on the work we did at our most recent residency, this is not correct. I think that @haithamshahin333 and I will have to make some adjustments.

InfoSec812:
This is going to be superseded by the work I am putting into #219

pabrahamsson:
@haithamshahin333 & @InfoSec812 based on your comments about this PR being superseded by #219, can we close this PR?

pabrahamsson:
Closing in favor of #219

PR #205, opened 2019-04-18
changed sonar Dockerfile to take plugin argument with defaults set

bvkin:
#### What is this PR About?
This changes the sonarqube Dockerfile to take a list of arguments for plugins.

#### How do we test this?
Run the openshift-applier inventory against a cluster. Ensure that the ARG step is present in the build.

### DIFF
--- a/sonarqube/Dockerfile
+++ b/sonarqube/Dockerfile
@@ -1,6 +1,7 @@
 FROM docker.io/sonarqube:latest
 
 USER root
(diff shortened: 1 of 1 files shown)


pcarney8:
after running the applier to test, it looks like this guy is missing the secret `sonardb` when trying to deploy it failes. That is probably from before, but the build runs to completion

pcarney8:
@bvkin approving. I will fix the template and add in the secret so it can work with the automation in the next PR

PR #206, opened 2019-04-18
put database service name in so sonarqube will actually deploy

pcarney8:
#### What is this PR About?
fix the params so that sonarqube will actually deploy

#### How do we test this?
Run the applier in the sonarqube directory, follow the readme


bvkin:
Tested this and it works

PR #208, opened 2019-05-03
change default slave image tag to latest instead of v3.11, template s…

pcarney8:
…hould be generic as possible

#### What is this PR About?
Changing the default from v3.11 to latest for the `SLAVE_IMAGE_TAG`

#### How do we test this?
apply the applier


pcarney8:
@springdo @etsauer @pabrahamsson Could I get a review here? this is related to https://github.com/rht-labs/labs-ci-cd/pull/274

PR #217, opened 2019-06-13
Add ipa

tylerauerbeck:
#### What is this PR About?
Adds an applier to deploy ipa-server (Red Hat IdM) on openshift

#### How do we test this?
Replace the appropriate values in `.applier/group_vars/seed-hosts.yml` and then run: 
- `ansible-galaxy install -r requirements.yml -p roles`
- `ansible-playbook -i .applier/ roles/openshift-applier/playbooks/openshift-cluster-seed.yml`

cc: @redhat-cop/day-in-the-life @makentenza @oybed @pabrahamsson


ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,61 @@
...
+namespace: ipa
+app_name: ipa
+base_openshift_url: example.com 
+realm: "EXAMPLE.COM"
+
+params:
+  IPA_SERVER_SERVICE: "{{ app_name }}"
+  IPA_SERVER_IMAGE: "freeipa-server:rhel-7"
+  IPA_SERVER_HOSTNAME: "{{ app_name }}.apps.{{ base_openshift_url }}"
+  IPA_SERVER_INSTALL_OPTS: "-U -r {{ realm }} --setup-dns --no-forwarders --no-ntp"
```
oybed: Curious to know what the `--setup-dns` will do in this case. We normally host DNS externally to the OpenShift cluster, so having another DNS server running in here may potentially mean "out-of-sync" DNS ... The good news is that IdM may be the only one confused. Did you try without the `--setup-dns` flag? I suspect that the "install" may fail if it's not specified?

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,61 @@
...
+  IPA_ADMIN_PASSWORD: "Password123"
+  SERVICE_ACCOUNT_USEROOT: "{{ app_name }}-sa"
+  VOLUME_CAPACITY: 1G
+  TIMEOUT: 600
+
+openshift_cluster_content:
+- object: Environment Setup
+  content:
+  - name: Create Projects
+    template: "https://raw.githubusercontent.com/redhat-cop/openshift-templates/v1.4.4/project-requests/create-project.yml"
```
oybed: can you create inventory variables at top for the tag? i.e.: `openshift_templates_version: v1.4.5`

ipa-server/README.md
```diff
@@ -0,0 +1,25 @@
...
+* The permissions to grant an SA the anyuid scc.
+
+### Environment Setup
+
+1. Clone this repository: `git clone https://github.com/redhat-cop/containers-quickstarts`
+2. `cd containers-quickstarts/ipa-server`
+3. Run `ansible-galaxy install -r requirements.yml --roles-path=roles`
+4. Login to OpenShift: `oc login -u <username> https://console.example.com:8443`
+
+:heavy_exclamation_mark: freeipa-server container will run under a replica of `anyuid` SCC. You should ensure you are logged into the Cluster (step 4) with an user with privileges to create SecurityContextConstraints with the required scope
```
oybed: This is ok for now, but I'd like to see a follow-up PR where we better define a custom role/rolebinding that allows for this to run without needing to run under `anyuid`

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,61 @@
...
+    action: create
+    params_from_vars:
+      NAMESPACE: "{{ namespace }}"
+      NAMESPACE_DISPLAY_NAME: "{{ namespace }}"
+    tags:
+      - project
```
oybed: Can we prefix these tags with `ipa_`? - i.e.: `ipa_project`, `ipa_imagestream`, etc.

tylerauerbeck: Fixed.

ipa-server/.applier/group_vars/seed-hosts.yml
tylerauerbeck: Fixed

ipa-server/.applier/group_vars/seed-hosts.yml
tylerauerbeck: I'll need to test this out. My initial experience with removing some of these didn't go well, but I also tried pulling a few out at a time. Let me try just removing that flag and I'll circle back.

ipa-server/README.md
tylerauerbeck: Agreed. This is just in place for the initial commit. I think this is possible, but will require some untangling of their init script (and likely some buildconfig setup)

ipa-server/.applier/group_vars/seed-hosts.yml
oybed: ok - that could be done post-merge as well. Just curious if this was tried and what the outcome was.

ipa-server/README.md
oybed: Perfect

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,64 @@
...
+    params_from_vars:
+      NAMESPACE: "{{ namespace }}"
+      NAMESPACE_DISPLAY_NAME: "{{ namespace }}"
+    tags:
+      - ipa_project
+
+- object: Image Streams
+  content:
+  - name: "{{ app_name }}-streams"
+    file: "https://raw.githubusercontent.com/freeipa/freeipa-container/5e55ee8a62e7c51bf8c1805b33425f48e2d8f44e/freeipa-server-openshift-image.json"
```
pabrahamsson: Does it make sense to use something like `{{ freeipa_templates_version }}` to prepare for future updates?

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,64 @@
...
+    params_from_vars:
+      NAMESPACE: "{{ namespace }}"
+      NAME: "{{ params.SERVICE_ACCOUNT_USEROOT }}"
+    tags:
+      - ipa_serviceaccount
+
+- object: Deployments
+  content:
+  - name: ipa
+    template: "https://raw.githubusercontent.com/freeipa/freeipa-container/5e55ee8a62e7c51bf8c1805b33425f48e2d8f44e/freeipa-server-openshift.json"
```
pabrahamsson: Same as above

ipa-server/.applier/group_vars/seed-hosts.yml
tylerauerbeck: Done

ipa-server/.applier/group_vars/seed-hosts.yml
tylerauerbeck: @oybed I dug into this a little bit and end up with the same result (not good). May be best left for a future iteration of digging into the `ipa-server-install` command flags.

tylerauerbeck:
@oybed @makentenza So after trying to run this more than once (without deleting anything) the applier runs into an issue when processing the `scc`. I'm going to compare the template I'm using against the one that I know is used as part of the `gitlab` quoickstart because I don't think we see that problem there (and I believe they're nearly identical). Error I'm seeing is: 

```
failed: [localhost] (item={'oc_path': u''}) => {"changed": true, "cmd": "oc process  --local  -f https://raw.githubusercontent.com/redhat-cop/openshift-templates/v1.4.5/scc/project-run-anyuid-template.yml -n ipa  --param='NAMESPACE=ipa' --param='NAME=ipa'  --ignore-unknown-parameters | oc apply -n ipa -f - ", "delta": "0:00:01.078751", "end": "2019-06-18 23:56:04.277120", "failed_when_result": true, "msg": "non-zero return code", "oc_param_file_item": {"oc_path": ""}, "rc": 1, "start": "2019-06-18 23:56:03.198369", "stderr": "Error from server: unrecognized type: int32", "stderr_lines": ["Error from server: unrecognized type: int32"], "stdout": "", "stdout_lines": []}```

**Comment by makentenza** [comment]
So @tylerauerbeck for the error you are getting it looks like the template is not correct:

https://github.com/redhat-cop/openshift-templates/blob/master/scc/project-run-anyuid-template.yml#L23

That entry should be:

```priority: ${{PRIORITY_LEVEL}}```

Note the double brackets, as the template is expecting a non-string parameter. Will submit a quick fix for that.

**Comment by makentenza** [comment]
@tylerauerbeck this should solve the issue:

https://github.com/redhat-cop/openshift-templates/pull/32

File: ipa-server/.applier/group_vars/seed-hosts.yml [hunk 7]
```diff
@@ -0,0 +1,66 @@
...
+    params_from_vars:
+      NAMESPACE: "{{ namespace }}"
+      NAMESPACE_DISPLAY_NAME: "{{ namespace }}"
+    tags:
+      - ipa_project
+
+- object: Image Streams
+  content:
+  - name: "{{ app_name }}-streams"
+    file: "https://raw.githubusercontent.com/freeipa/freeipa-container/{{ ipa_templates_version }}/freeipa-server-openshift-image.json"
```
bvkin: This image stream seems to pull fedora/centos base images but up above we specify rhel-7:

```
$oc get is
NAME             DOCKER REPO                                           TAGS                                       UPDATED
freeipa-server   docker-registry.default.svc:5000/ipa/freeipa-server   centos-7,fedora-24,fedora-25 + 1 more...   2 minutes ago
```

bvkin: I guess the hash for `ipa_templates_version` doesn't include rhel. This one does though `ccf9229b9c43e9c9fa9a83ef70d1e6ade8e0f3fa`

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,66 @@
...
+    params_from_vars:
+      NAMESPACE: "{{ namespace }}"
+      NAME: "{{ params.SERVICE_ACCOUNT_USEROOT }}"
+    tags:
+      - ipa_serviceaccount
+
+- object: Deployments
+  content:
+  - name: ipa
+    template: "https://raw.githubusercontent.com/freeipa/freeipa-container/{{ ipa_templates_version }}/freeipa-server-openshift.json"
```
bvkin: There seems to be an issue with idempotency here. When I run the applier the first time it works. After that I get an error:
```
failed: [localhost] (item={'oc_path': u''}) => {"changed": true, "cmd": "oc process  --local  -f https://raw.githubusercontent.com/freeipa/freeipa-container/504c122090713505025971cee7a4d50eb394e291/freeipa-server-openshift.json -n bkin-ipa  --param='IPA_SERVER_HOSTNAME=ipa.apps.example.com' --param='IPA_SERVER_IMAGE=freeipa-server:rhel-7' --param='VOLUME_CAPACITY=1G' --param='IPA_SERVER_INSTALL_OPTS=-U -r EXAMPLE.COM --setup-dns --no-forwarders --no-ntp' --param='SERVICE_ACCOUNT_USEROOT=ipa' --param='TIMEOUT=600' --param='IPA_ADMIN_PASSWORD=password'  --ignore-unknown-parameters | oc apply -n bkin-ipa -f - ", "delta": "0:00:00.819589", "end": "2019-06-25 11:30:32.229980", "failed_when_result": true, "msg": "non-zero return code", "oc_param_file_item": {"oc_path": ""}, "rc": 1, "start": "2019-06-25 11:30:31.410391", "stderr": "Error from server (Invalid): Service \"freeipa-server\" is invalid: spec.clusterIP: Invalid value: \"\": field is immutable\nError from server: unrecognized type: int64", "stderr_lines": ["Error from server (Invalid): Service \"freeipa-server\" is invalid: spec.clusterIP: Invalid value: \"\": field is immutable", "Error from server: unrecognized type: int64"], "stdout": "route.route.openshift.io/freeipa-server-https unchanged\nsecret/freeipa-server-password configured\npersistentvolumeclaim/freeipa-server unchanged", "stdout_lines": ["route.route.openshift.io/freeipa-server-https unchanged", "secret/freeipa-server-password configured", "persistentvolumeclaim/freeipa-server unchanged"]}
```

tylerauerbeck: Yepp, I've noticed that as well. Trying to see if I can get around that without having to maintain our own version of that template.

ipa-server/.applier/group_vars/seed-hosts.yml
tylerauerbeck: Actually it looks like it does, if you look at the file here: https://raw.githubusercontent.com/freeipa/freeipa-container/5e55ee8a62e7c51bf8c1805b33425f48e2d8f44e/freeipa-server-openshift-image.json , you'll see `rhel-7` at the bottom.

bvkin: Oh woops, looks like I had a local change. Not using hashes is rough.

ipa-server/requirements.yml
```diff
@@ -0,0 +1,8 @@
+# This is the Ansible Galaxy requirements file to pull in the correct roles
+# to support the operation of CASL provisioning/runs.
+
+# From 'casl-ansible'
+- name: openshift-applier
+  scm: git
+  src: https://github.com/redhat-cop/openshift-applier
+  version: v2.0.10
```
oybed: Please update to v2.1.1

tylerauerbeck: @oybed I'm actually seeing some issues when changing from v2.0.10 to v.2.1.1. Probably something small, but not sure I'll have time to resolve before I leave. Could we look at leaving at the current version and then resolving in a follow-up PR?

tylerauerbeck:
@pabrahamsson @oybed Updated and now runs correctly on both OCP v3 and v4 (as long as settings as noted in the README are followed). Feel free to take a look.

ipa-server/requirements.yml
oybed: sounds good

PR #233, opened 2019-07-25
Ensure tests fail if there are no builds found

etsauer:
#### What is this PR About?
We want to make sure we don't get false positives on tests where we're looking in the wrong namespace. To that, the test script now expects to find at least one build.

#### How do we test this?

```
oc login ...
./_test/setup.sh applier esauer-test-scripts etsauer/containers-quickstarts test-script-error && ./_test/setup.sh test esauer-test-scripts etsauer/containers-quickstarts test-script-error
```

### DIFF
--- a/_test/setup.sh
+++ b/_test/setup.sh
@@ -1,13 +1,9 @@
 #!/bin/bash
 trap "exit 1" TERM
 export TOP_PID=$$
-<<<<<<< Updated upstream
-NAMESPACE=containers-quickstarts-tests
-=======
 NAMESPACE="${2:-containers-quickstarts-tests}"
 TRAVIS_REPO_SLUG="${3:-redhat-cop/containers-quickstarts}"
 TRAVIS_BRANCH="${4:-master}"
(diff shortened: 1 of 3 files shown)


_test/setup.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
 trap "exit 1" TERM
 export TOP_PID=$$
+<<<<<<< Updated upstream
 NAMESPACE=containers-quickstarts-tests
+=======
+NAMESPACE="${2:-containers-quickstarts-tests}"
+TRAVIS_REPO_SLUG="${3:-redhat-cop/containers-quickstarts}"
+TRAVIS_BRANCH="${4:-master}"
+>>>>>>> Stashed changes
```
pabrahamsson: Looks like you had a merge conflict here. Can you please clean up?

etsauer:
@pabrahamsson fixed some issues, and enhanced the travis config to make use of the new parameters.

PR #234, opened 2019-07-26
Add new Docker build templates to allow overriding base images

### LINKED ISSUE #186: Pin SonarQube version for each release
Currently, the SonarQube image is always the "latest" version of SonarQube, thus there is no guarantee that any given release of this project will have the same version of SonarQube.

From https://github.com/rht-labs/labs-ci-cd/issues/64

InfoSec812:
#### What is this PR About?
This PR adds a new Docker buildconfig template which allows for overriding the `FROM` clause of the target Dockerfile build. This will support resolving #186 in [Labs CI/CD](https://github.com/rht-labs/labs-ci-cd)

#### How do we test this?
Apply this template with openshift-applier to provision SonarQube and provide the DockerImage and DockerImageTag and ensure that the correct instance of SonarQube is used as the base image.

Resolves #186


InfoSec812:
> Thank you @InfoSec812
> Might I suggest adding this template to `redhat-cop/openshift-templates` instead and reference it from there in [seed-hosts.yml](/redhat-cop/containers-quickstarts/blob/master/build-docker-generic/.applier/group_vars/seed-hosts.yml)?

@pabrahamsson We could, but this template is specifically required for SonarQube so that we can pin to version 7.7, and my understanding is that things that are specific to a particular container image we would want here.

pabrahamsson:
@InfoSec812 this looks like a great example of a generic template to me. Can you just help me understand what it is that is specific to SonarQube? Maybe I'm missing some context here.

InfoSec812:
@pabrahamsson Well, my reasoning is that it is being created to support a specific need for SonarQube and that need has not arose anywhere else as yet that I am aware of.

tylerauerbeck:
@InfoSec812 I think we discussed this recently on the CoP call. So while yes, this fixes the issue that we're specifically having with SQ here.. this could be useful in general to a number of other use cases. So I think in this instance, this new template probably is better served living in `redhat-cop/openshift-templates`

InfoSec812:
> @InfoSec812 I think we discussed this recently on the CoP call. So while yes, this fixes the issue that we're specifically having with SQ here.. this could be useful in general to a number of other use cases. So I think in this instance, this new template probably is better served living in `redhat-cop/openshift-templates`

@tylerauerbeck As discussed on the CoP call, I am not going to have time to do that in the near future, so if someone could help I'm sure many would appreciate it.

tylerauerbeck:
@InfoSec812 Must have missed that. Sorry! I can take a look at helping out there if that's the case.

PR #235, opened 2019-07-29
WIP: Use ETL CI cluster

pabrahamsson:
#### What is this PR About?
Move CI run to ETL CI cluster since we're hitting resource limits with `cluster up` in Travis.
Also updating the oc client to 3.11.

#### How do we test this?
Should kick off automatically as part of our Travis CI integration.
Might be issues with the way we're trying to use the OCP token, see [Travis documentation](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,12 @@ cache:
 
 env:
   global:
-    - ANSIBLE_HOST_KEY_CHECKING=False
-    - PIP_DOWNLOAD_CACHE=$HOME/.cache/pip
-    - OC_BINARY_URL=https://mirror.openshift.com/pub/openshift-v3/clients/3.11.133/linux/oc.tar.gz
+    - ANSIBLE_HOST_KEY_CHECKING="False"
+    - PIP_DOWNLOAD_CACHE="$HOME/.cache/pip"
+    - OC_BINARY_URL="https://mirror.openshift.com/pub/openshift-v3/clients/3.11.133/linux/oc.tar.gz"
     - ANSIBLE_VERSION="2.7"
-    - CI_CLUSTER_URL=https://console.ci-1.cop.rht-labs.com
+    - OCP_CLUSTER_URL="https://console.ci-1.cop.rht-labs.com"
+    - LE_CA_CERT="https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem.txt"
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,10 +9,12 @@ cache:
...
 env:
   global:
-    - ANSIBLE_HOST_KEY_CHECKING=False
-    - PIP_DOWNLOAD_CACHE=$HOME/.cache/pip
-    - OC_BINARY_URL=https://mirror.openshift.com/pub/openshift-v3/clients/3.10.45/linux/oc.tar.gz
+    - ANSIBLE_HOST_KEY_CHECKING="False"
+    - PIP_DOWNLOAD_CACHE="$HOME/.cache/pip"
+    - OC_BINARY_URL="https://mirror.openshift.com/pub/openshift-v3/clients/3.11.133/linux/oc.tar.gz"
     - ANSIBLE_VERSION="2.7"
+    - OCP_CLUSTER_URL="https://console.ci-1.cop.rht-labs.com"
```
oybed: Would prefer to not add our URLs in public github, so let's see if we can find a way to pass this in without keeping it in the repo(s).

.travis.yml
```diff
@@ -9,10 +9,12 @@ cache:
 
 env:
   global:
-    - ANSIBLE_HOST_KEY_CHECKING=False
-    - PIP_DOWNLOAD_CACHE=$HOME/.cache/pip
-    - OC_BINARY_URL=https://mirror.openshift.com/pub/openshift-v3/clients/3.10.45/linux/oc.tar.gz
+    - ANSIBLE_HOST_KEY_CHECKING="False"
+    - PIP_DOWNLOAD_CACHE="$HOME/.cache/pip"
+    - OC_BINARY_URL="https://mirror.openshift.com/pub/openshift-v3/clients/3.11.133/linux/oc.tar.gz"
```
oybed: Good change. I know there was issues with OC client v3.11.x for earlier CI runs and it has been holding us back from moving to v3.11, but hopefully those issues have been resolved now (wonder if it was an issue with Travis CI env actually and hence we should be good when moving away).

pabrahamsson:
Closing as we are moving away from using Travis CI with our own OCP instance.

PR #219, opened 2019-06-20
Update ZAP Dockerfile

InfoSec812:
#### What is this PR About?
This PR updates the containers-quickstarts for a reliable ZAP implementation. The Jenkins Agent Dockerfile.rhel7 works well even when using Multi-container Jenkins Pods. The Dockefile.rhel7-daemon file allows you to deploy ZAP as a persistent service which can then be used as needed from any pipeline.

#### How do we test this?
Check the `README.md` and the `Multi-Container_Pipeline_In_Jenkins_On_OpenShift.md` in the `jenkins-slaves/jenkins-slave-zap` directory.

cc: @redhat-cop/day-in-the-life

PR #237, opened 2019-08-03
UPDATE - make the zap slave actually do stuff

### LINKED ISSUE #236: zap scanner does not work cuz all the .py files are really old and some other reasons too
We have been trying to update the DO500 course content and hit several issues with the zap slave. The image builds fine, but errors like crazy when you try and run a scan. 

I am working on a "🔨fix 🔨" for this, it's a bit .... interesting .... but should be along shortly

springdo:
#### What is this PR About?
fixes #236 236

#### How do we test this?
You can download the code base, run a build of the image, have jenkins, have an app you want to test against in a jenkins pipeline, and generate a html report. 
If the html report won't generate, then this will fail 🤷‍♂ 


springdo:
@oybed @eformat @tylerauerbeck - we're needing this to progress with scanning exercise in DO500 land.

InfoSec812:
There is a reliable and tested configuration awaiting approval in #219. Why are we re-doing all of this work? @springdo @etsauer ?

etsauer:
@InfoSec812 didn't realize these were doing similar things. I'll take a look and comment over there.

PR #242, opened 2019-08-12
Updated default context dir value as ansible-stacks no longer exists

garethahealy:
#### What is this PR About?
The default context dir value does not exist. Directory without *-stacks exists.

#### How do we test this?
Process yaml and apply


garethahealy:
- https://github.com/redhat-cop/containers-quickstarts/commit/60a5b004fd8ab8723cff775da33f7146c5f72360

PR #250, opened 2019-09-03
Fix sonarqube builds and use latest release

pabrahamsson:
#### What is this PR About?
The latest upstream release of Sonarqube will not run on Openshift. This PR will let you disable memory mapping for Elasticsearch and let Sonarqube start.
It will also let you specify the upstream version of Sonarqube and thus you can revert to `7.7-community` release if you don't want to disable memory mapping in Elasticsearch.

#### How do we test this?
Follow instructions in README


InfoSec812:
This is FANTASTIC. Thanks @pabrahamsson APPROVED!

PR #260, opened 2019-10-01
Issue 259: Nexus OpenShift plugin

InfoSec812:
#### What is this PR About?
This provides a configuration for deployment of Nexus repository manager along with the OpenShift/Kubernetes configuration plugin

#### How do we test this?
1. Install Ansible on your local machine
1. Log on to the OpenShift cluster with the CLI
1. Import the `sonatype/nexus-repository-manager` image as an ImageStream into the `openshift` namespace
   * `oc import-image sonatype/nexus-repository-manager --from=registry.connect.redhat.com/sonatype/nexus-repository-manager --confirm`
2. Clone this repository and open a terminal in the root of the project
3. Change directory to `./nexus/test`
4. Install the OpenShift Applier role(s)
   1. `ansible-galaxy install -r requirements.yml -p ./roles`
5. Run the applier playbook
   1. `ansible-playbook -i inventory test.yml`
6. Wait for the Nexus application to become available
7. Open the Nexus page and attempt to log on as `admin` with password `thisismytestpassword`
   1. If this works, the container is correct.


nexus/test/apply.retry
```diff
@@ -0,0 +1 @@
+localhost
```
pabrahamsson: Please remove this file

nexus/test/inventory/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,62 @@
...
+  deploy:
+    VOLUME_CAPACITY: 10Gi
+    MEMORY_LIMIT: 2Gi
+    CONTAINER_IMAGE: sonatype/nexus3:latest
+
+openshift_cluster_content:
+- object: Environment Setup
+  content:
+  - name: Create Projects
+    template: "https://raw.githubusercontent.com/redhat-cop/cluster-lifecycle/v3.9.0/files/projectrequest/template.yml"
```
pabrahamsson: Can we switch to use `{{ openshift_templates_raw }}/{{ openshift_templates_raw_version }}/project-requests/create-project.yml`?

nexus/test/requirements.yml
```diff
@@ -0,0 +1,8 @@
+# This is the Ansible Galaxy requirements file to pull in the correct roles
+# to support the operation of CASL provisioning/runs.
+
+# From 'openshift-applier'
+- name: openshift-applier
+  scm: git
+  src: https://github.com/redhat-cop/openshift-applier
+  version: v2.0.8
```
pabrahamsson: Please update to `v2.1.1` so this also works on OCP 4

nexus/test/inventory/group_vars/seed-hosts.yml
```diff
@@ -27,7 +27,7 @@ openshift_cluster_content:
 - object: Environment Setup
   content:
   - name: Create Projects
-    template: "https://raw.githubusercontent.com/redhat-cop/cluster-lifecycle/v3.9.0/files/projectrequest/template.yml"
+    template: "{{ openshift_templates_raw }}/{{ openshift_templates_raw_version }}/projectrequest/template.yml"
```
pabrahamsson: Please see my original comment regarding the url as this one will give you 404.

nexus/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM image-registry.openshift-image-registry.svc:5000/openshift/nexus-repository-manager
```
pabrahamsson: Can you please verify that this works on both OCP 3 & 4?
If this only works on 4 then I think we need to be explicit about that in the README.

garethahealy:
@InfoSec812 few comments.

1. Looking at the structure, all the applier related ansible lives under /test, where as the standard location currently is .applier; see: https://github.com/redhat-cop/containers-quickstarts/tree/master/build-s2i-executable - would be better to follow whats already been done.
2. You've included a template but upstream already provides one here: https://github.com/sonatype/docker-rhel-nexus/blob/master/OpenShift/nexus-rhel.json - should we not use the nexus upstream version?
3. As @pabrahamsson mentioned, image uses a svc name from OCP4 which wont work on OCP3 by default

PR #264, opened 2019-10-23
Add OWNERS file

pabrahamsson:
#### What is this PR About?
Add OWNERS file (see [documentation](https://go.k8s.io/owners)) to leverage Prow plugins

#### How do we test this?

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -14,5 +14,3 @@ filters:
       - pcarney8
       - pabrahamsson
       - sabre1041
-
-- should error
--- a/OWNERS
+++ b/OWNERS
@@ -1,16 +1,16 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
-filters:
-  ".*":
-    reviewers:
-      - etsauer
-      - oybed
-      - pcarney8
-      - pabrahamsson
-      - sabre1041
-    approvers:
-      - etsauer
-      - oybed
-      - pcarney8
-      - pabrahamsson
-      - sabre1041
+reviewers:
+  - etsauer
+  - oybed
+  - pcarney8
+  - pabrahamsson
+  - sabre1041
+  - springdo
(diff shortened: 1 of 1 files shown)


pabrahamsson:
@etsauer yes we can as long as this comment in the [documentation](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners_aliases) isn't a concern for us.
>We use aliases for groups instead of GitHub Teams, because changes to GitHub Teams are not publicly auditable.

etsauer:
@pabrahamsson hmm, i guess i don't know what a "group" is then. I thought they were the same thing. @sabre1041 ?

pabrahamsson:
@etsauer I think what they're saying is that they create their own "groups" in the `OWNERS_ALIASES` file instead of using GH teams.

sabre1041:
@etsauer @pabrahamsson it depends. There needs to be an understanding that groups are distinct from GH teams and that they would need to be managed separately

etsauer:
@sabre1041 but that's what I am trying to understand... HOW do you manage a group? I don't know what that is, or how its different from a team.

pabrahamsson:
My understanding of a group (`core-group`):
```
aliases:
  core-group:
    - user1
    - user2
```
This is managed by the `OWNERS_ALIASES` file and you/we need to manually update it through PRs.
The `core-group` can now be used in `OWNERS` files in the repo.

sabre1041:
@etsauer groups are a prow convention that is managed through files within the repository where as teams are a GitHub convention

pabrahamsson:
@etsauer @sabre1041 are we good as is? We can always explore using GH teams and/or `OWNERS_ALIASES` down the road.

etsauer:
@pabrahamsson yeah, i'm good with it. Let's merge and address it when it becomes a problem.

PR #273, opened 2019-11-14
Update ipa quickstart for OCP4

tylerauerbeck:
#### What is this PR About?
In order to run the ipa quickstart on OCP4, a certain sebool needs to be set. This PR does some cleanup on this quickstart and adds a machineconfig that puts this sebool in place.

#### How do we test this?
On OCP4 -- run the following:
```
ansible-playbook -i .applier/ roles/openshift-applier/playbooks/openshift-cluster-seed.yml -e 'include_tags=ipa_machineconfig'
```
You can then confirm that the machineconfig has been put in place by monitoring your nodes and machineconfigpools with

```
oc get nodes
oc get machineconfigpools
```
When looking at the machineconfigpools you can see that you are ready to proceed when the worker pool indicates that it has been updated and that updating status is set to false (i.e. no longer updating). For your nodes, you'll then see that your worker nodes are ready and schedulable again.

Once this is in place, you can then run the rest of the applier as expected:

```
ansible-playbook -i .applier/ roles/openshift-applier/playbooks/openshift-cluster-seed.yml -e 'exclude_tags=ipa_machineconfig'
```


ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -44,7 +49,7 @@ openshift_cluster_content:
     template: "https://raw.githubusercontent.com/redhat-cop/openshift-templates/{{ openshift_templates_version }}/service-accounts/generic-service-account.yml"
     namespace: "{{ namespace }}"
```
pabrahamsson: This is not necessary since we're setting the namespace in the template

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -53,7 +58,7 @@ openshift_cluster_content:
...
```
pabrahamsson: I don't think SCCs are namespaced so this can be removed

ipa-server/.applier/group_vars/seed-hosts.yml
```diff
@@ -29,6 +29,11 @@ openshift_cluster_content:
       NAMESPACE_DISPLAY_NAME: "{{ namespace }}"
     tags:
       - ipa_project
+  - name: Deploy MachineConfig to configure SEbool
+    file: "{{ inventory_dir }}/../.openshift/files/01-worker-sebool.yml"
+    tags:
+      - ipa_ocp4
+      - ipa_machineconfig
```
pabrahamsson: could we use the `never` tag here? This way the installation instructions can be simplified such that the second step (deployment) doesn't need any tags (include or exclude).

tylerauerbeck:
@pabrahamsson Requested changes have been made.

PR #283, opened 2020-01-06
Change how JVM properties are applied to allow more flexible configuration

### LINKED ISSUE #282: Re-order options so that when env var is undefined it will not fail the script
Having the `${SONARQUBE_WEB_JVM_OPTS}` BEFORE the other options means that if that env is not set, then the application will fail to start because of the extra space between the `=` and the next parameter. We should re-order so that the static options are before the environment var and then the problem will be resolved.

https://github.com/redhat-cop/containers-quickstarts/blob/ecb424d69c95a86d7f26570d7cc62b6c3b2d22fb/sonarqube/run.sh#L27

InfoSec812:
#### What is this PR About?
When the environment variable `SONARQUBE_WEB_JVM_OPTS` is undefined, the application will not start because there is an extra space between the property name and the next property value.

#### How do we test this?

1. get MR code on your machine

1. following instructions in [README](https://github.com/redhat-cop/containers-quickstarts/tree/master/sonarqube)

1. then, since this repo will point the build to the master branch, run a new build

```oc start-build sonarqube --from-dir=.. -n sonarqube```

1. Go to sonar home page. There should now be a 'login with OpenShift link'
1. You should be redirected to the OpenShift login screen. --> Login.
1. You should be redirected to Sonarqube and see the projects page (empty)

Resolves #282 

### DIFF
--- a/sonarqube/.openshift/templates/sonarqube-deployment-template.yml
+++ b/sonarqube/.openshift/templates/sonarqube-deployment-template.yml
@@ -116,6 +116,10 @@ objects:
                 name: sonar-ldap-bind-dn
           - name: SONAR_SEARCH_JAVA_ADDITIONAL_OPTS
             value: '-Dnode.store.allow_mmapfs=false'
+          - name: SONAR_SAR_ADMIN_GROUP
+            value: ${SONAR_SAR_ADMIN_GROUP}
+          - name: SONAR_SAR_USER_GROUP
+            value: ${SONAR_SAR_USER_GROUP}
           imagePullPolicy: Always
           livenessProbe:
             failureThreshold: 3
@@ -295,3 +299,11 @@ parameters:
   - name: SONAR_SEARCH_JAVA_ADDITIONAL_OPTS
     description: Pass in additional Java opts to ElasticSearch
     displayName: Add sonar.search.javaAdditionalOpts
+  - name: SONAR_SAR_ADMIN_GROUP
+    description: Group which will be given admin privileges in SonarQube
+    displayName: OAuth Admin Group
+    value: sonar-administrators
+  - name: SONAR_SAR_USER_GROUP
+    description: Group which will be given user privileges in SonarQube
+    displayName: OAuth User Group
+    value: sonar-users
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by pabrahamsson** [comment]
Still WIP @InfoSec812?

**Comment by InfoSec812** [comment]
@pabrahamsson Yes, still WIP... I should have a final solution in an hour or so.

File: sonarqube/.openshift/templates/sonarqube-deployment-template.yml [hunk 1]
```diff
@@ -185,8 +197,8 @@ objects:
       name: sonarqube
       weight: 100
     wildcardPolicy: None
-- apiVersion: v1
-  kind: Service
+- kind: Service
+  apiVersion: v1
```
pabrahamsson: Same as above

sonarqube/.openshift/templates/sonarqube-deployment-template.yml
```diff
@@ -201,14 +213,14 @@ objects:
       deploymentconfig: sonarqube
     sessionAffinity: None
     type: ClusterIP
-- apiVersion: v1
-  kind: ServiceAccount
+- kind: ServiceAccount
+  apiVersion: v1
```
pabrahamsson: Same as above

sonarqube/.openshift/templates/sonarqube-deployment-template.yml
```diff
@@ -201,14 +213,14 @@ objects:
...
   metadata:
     annotations:
       serviceaccounts.openshift.io/oauth-redirectreference.sonarqube: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"sonarqube"}}'
     name: sonarqube
-- apiVersion: v1
-  kind: RoleBinding
+- kind: RoleBinding
+  apiVersion: v1
```
pabrahamsson: Same as above

pabrahamsson:
@InfoSec812 this PR only need a few formatting changes and it will be good to go. Do you think you'll be able to take care of this in the near future?

Issue #276, opened 2019-12-13
Update tool-box container with Openshift 4 toolset


huddlesj:
An updated tool-box container is needed including the Openshift 4 clients and maybe all the new container build and management tools ( podman, buildah, crictl ).

huddlesj:
An updated tool-box container is needed including the Openshift 4 clients and maybe all the new container build and management tools ( podman, buildah, crictl ).

PR #293, opened 2020-01-30
Updated to oc 4.3 client

huddlesj:
Updated OC client for issue #276 

#### What is this PR About?
Updated oc client in default Docker file. Also as example added an OC3 only and a oc3/oc4 Dockerfile

#### How do we test this?

Use oc client to connect to Openshift 4 cluster and oc3 client to connect to an Openshift 3 cluster.


garethahealy:
is this needed? shouldn't the 4.3 client be backwards compatible with 3.11?

huddlesj:
@garethahealy I wasn't sure if we could get rid of the 3 client or if we would need to maintain it in a separate tag, branch or release. If it is not needed I can git rid of the oc3 and the multi Docker files.

etsauer:
@huddlesj @garethahealy yeah, I think we can remove these new images. we have older tags available on quay that will still have the 3.11 oc version if people need it. https://quay.io/repository/redhat-cop/tool-box?tab=tags

pabrahamsson:
Still WIP @huddlesj? Or are we good to merge?

huddlesj:
@pabrahamsson Yes, We are good to merge.

PR #307, opened 2020-03-31
Various UX improvements and bug fixes to eap72 chart.

deweya:
#### What is this PR About?
Improving EAP72 Helm chart by adding ux enhancements and bug fixes.

Changes include:
- Changing jgroups keystore from pkcs12 to jceks (was causing "Invalid keystore format" errors)
- Disabling bc triggers (was causing a confusing user experience because the initial installation was auto-triggering, but upgrades were not)
- Grouping all tls-related values under a `tls` map
- Adding tls.httpsSecret value (default to https-keystore)
- Saving https keystore secret in OCP secret
- Changing default keystore password to "changeit"
- Various README edits

#### How do we test this?
"helm test" command after installing the chart to an ocp cluster

### DIFF
--- a/helm-charts/eap72/README.md
+++ b/helm-charts/eap72/README.md
@@ -1,8 +1,8 @@
 # EAP 7.2 Helm Chart
-A Helm chart for deploying an EAP application to OpenShift.
+A Helm chart for building and deploying an EAP application to OpenShift, inspired by the [OpenShift EAP templates](https://github.com/jboss-openshift/application-templates/tree/master/eap).
 
 ## Installation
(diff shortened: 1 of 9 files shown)


pabrahamsson:
Thank you for this contribution @deweya, what is the proper way to test this? Do you already have a workflow you're using?
/approve

deweya:
To test, I deployed the chart to a cluster and ran the "helm test" command. The "helm test" command creates a Pod that attempts a curl against the deployed app.

In this PR, I ran the test command twice - once when deploying the chart with tls.enabled=false and another with tls.enabled=true. In a later PR, there could be an enhancement that uses [Chart Testing](https://github.com/helm/chart-testing) and GitHub Actions to automatically test PRs related to Helm charts.

PR #308, opened 2020-04-16
ubi8-git and ubi8-google-api-python-client - init add

itewk:
#### What is this PR About?
Adds two utility images useful for use in GitLab CI pipelines, but could be used in other places to.

#### How do we test this?
You could build the docker image and run it if you like. We are using these images in an internal pipeline at the moment but thought it best to share them with the world.

### DIFF
--- a/ubi8-git/.openshift/bc-ubi8-git.yml
+++ b/ubi8-git/.openshift/bc-ubi8-git.yml
@@ -1,0 +1,53 @@
+---
+apiVersion: v1
+kind: Template
+labels:
+  template: ubi8-git
+metadata:
+  annotations:
+    description: Git utility image
+    tags: git, ubi8
+  name: ubi8-git
+objects:
+- apiVersion: "v1"
+  kind: "BuildConfig"
+  metadata:
+    name: "ubi8-git"
+  spec:
+    output:
+      to:
+        kind: "ImageStreamTag"
+        name: "ubi8-git:latest"
+    source:
+      contextDir: "${CONTEXT_DIR}"
+      git:
+        uri: "${SOURCE_REPOSITORY_URL}"
+        ref: "${SOURCE_REPOSITORY_REF}"
+    strategy:
+      dockerStrategy:
+        dockerfilePath: Dockerfile
+    resources:
+      requests: 
+        cpu: 1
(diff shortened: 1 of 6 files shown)


itewk:
@etsauer would appreciate your review and merge here. trying to upstream some stuff we are using internally. with hopes of getting it into quay.io/redhat-cop.

itewk:
@pabrahamsson could i get you to look at this one too, its same idea as https://github.com/redhat-cop/containers-quickstarts/pull/309

itewk:
@pabrahamsson looks like it needs a `lgtm` from you to merge.

pabrahamsson:
It does @itewk but I need a bit more time on this one before I'm ready with the `lgtm`

itewk:
kk.

itewk:
@pabrahamsson okay. if we come up with any more images used for gitlab runners we will make a a parrent folder for them.

itewk:
@pabrahamsson does something special have to happen to have these images built and pushed to https://quay.io/organization/redhat-cop?

pabrahamsson:
AFAIK know it is not something that happens by default. I'm finding out what the process is to have this done.

pabrahamsson:
@itewk turns out the process to build images from a monorepo (such as this one) doesn't really exist. I started some work to make this possible (through GH actions) that I hope to have available shortly (currently in review).

itewk:
@pabrahamsson i thought for sure @etsauer had it set up to build and push the images automatically to quay.io/redhat-cop. I remember the “automagic” for the Jenkins workers images being busted at one point and he fixed it.

etsauer:
@itewk you're right, we did, but it was all set up in builds on the quay side that I was having to manually couch sound the quay ui to set up for each image. We wanted to switch to something more templated and repeatable using GitHub workflows that contributors could implement themselves.

itewk:
@etsauer anything can do to help? not sure how github workflows work, but if its anything like Gitlab CI and you need a semi stable OCP cluster to run things on....i may have one of those.

in the mean time while the "automatic" gets figured out, is there a manual way we could get these images and the one from https://github.com/redhat-cop/containers-quickstarts/pull/309 built and pushed to the quay.io/redhat-cop space?

PR #314, opened 2020-04-22
ubi8-git and ubi8-google-api - update to use registry.access.redhat.com/ubi8

itewk:
#### What is this PR About?
Fixes the source for the ubi8 image to be explicate so it can build in quay.io

#### How do we test this?
example: https://quay.io/repository/itewk/ubi8-git?tab=builds


itewk:
@pabrahamsson thanks for the quick merge

PR #320, opened 2020-04-24
Update rabbitmq configuration

gil0109:
#### What is this PR About?
Update to latest version of RabbitMQ (3.8.3) as default build
Add RabbitMQ HA policy to sync between pods
Add Pod AntiAffinity to ensure that all pods are on different nodes
Set minimum resource limits to satisfy additional HA load
Updated Documentation in configmap entry

#### How do we test this?
1. Deploy as you normally do.
2. Attach connect to an application that uses the default queue "/"
3. Run this command in terminal to ensure the queue is being synced between three pods
    rabbitmqctl list_queues name slave_pids synchronised_slave_pids
 
### Other Notes:
I deployed in my Openshift instance and confirmed working.


gil0109:
I am sorry but i think that over my head.

I would love to see what you did to learn though.  

Regards,
Karim

PR #325, opened 2020-04-27
🦆 FIX - Jenkins slave npm chrome dep 🦆

springdo:
#### What is this PR About?
As per discussion with @oybed and @pabrahamsson  - I've removed the `ADD` in favour of `curl` to bring in the Chrome Browser 

#### How do we test this?
`docker build -t name/jenkins-slave-npm .`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,11 @@ A set of buildConfigs for building custom Jenkins images for OpenShift.
 ## Related Content
 
 * [Container Pipelines](https://github.com/redhat-cop/container-pipelines) - A set of Jenkins piplines for OpenShift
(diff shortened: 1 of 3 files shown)


garethahealy:
@springdo ; instead of curl'ing, can we use chromium? it's part of the RH yum repos

springdo:
The rhel image Dockerfile uses a proper repo for chrome, but yeah if there was one for centos using chromium we could swap to it?

pabrahamsson:
@springdo @garethahealy ideally we'd like to use one of the upstream [UBI based jenkins images](https://quay.io/openshift/origin-jenkins-agent-base) but I know having to include a browser makes that challenging.

garethahealy:
It's in epel (if my memory is correct) - so you don't have to include other yum repos, i.e.: the google one which says they don't support RHEL

PR #330, opened 2020-05-01
WIP: add github workflows for gitlab-runners/ubi8-asciidoctor

itewk:
### WARNING
Before merging be sure to create the following repos:

* `quay.io/redhat-cop/gitlab-runners-ubi8-asciidoctor`

### What is this PR About?
* add github workflows for gitlab-runners/ubi8-asciidoctor
* gitlab-runners/ubi8-asciidoctor/Dockerfile - remove duplicate git install

#### How do we test this?
You can see https://github.com/itewk/containers-quickstarts/actions for example of workflows running.


itewk:
# WARNING #

Before merging be sure to create the following repos:

    quay.io/redhat-cop/gitlab-runners-ubi8-asciidoctor

itewk:
@buuhsmead FYI, this will make it so the asciidoctor image gets published to quay.io/redhat-cop/gitlab-runners-ubi8-asciidoctor

itewk:
though now that i think about it the ubi8-git and ubi8-asciidocotor images arn't gitlab runner specific at all....so probalby they shouldn't be in this folder....a different MR to fix that maybe?

itewk:
@pabrahamsson or maybe shoudl fix the pathing on this image before pushing to quay.io....

buuhsmead:
Not sure what you are doing with this image. Also not sure if adding gitlab-runners-code to this image will render it specific to gitlab. And not usable for plain podman (docker) runs anymore.

itewk:
@buuhsmead  didn't add any gitlab-runner specific code, but when i reorganized images in a different MR i accidnetly without thinking group them all under a gitlab-runners folder since that was the contexct they were getting used in. really this and the ubi8-git image should just be moved to a "utils" folder.....sigh. another refactor

itewk:
replaced by https://github.com/redhat-cop/containers-quickstarts/pull/331

Issue #313, opened 2020-04-20
Use bats for testing certain functionality within a built image


garethahealy:
As per discussion:
- https://github.com/redhat-cop/containers-quickstarts/pull/312#discussion_r411171014

PR #334, opened 2020-05-06
Add initial Nexus helm chart

tylerauerbeck:
#### What is this PR About?
This adds an initial helm chart to deploy Sonatype Nexus. This instance will include the Nexus Kubernetes plugin and will deploy the Red Hat and JBoss Maven repositories by default. This builds off of an upstream chart by pulling it in as a dependency while we work with the upstream chart owners to merge in our changes. Eventually we'll be able to remove our custom work here and just have instructions that say "use this upstream chart with the provided values file in this repo".

#### How do we test this?
From the `nexus/chart/nexus` directory, run:
```
oc new-project nexus-test
helm dep up
helm install nexus .         (<-- notice the trailing dot)
``` 

Once that's been deployed, you should see a running Nexus instance and can login with the default admin credentials of `admin:admin123)`

### DIFF
--- a/nexus/README.md
+++ b/nexus/README.md
@@ -1,0 +1,5 @@
+# Nexus
+
+This repository provides the following deployment strategies for Sonatype Nexus:
+
+- Helm: A Helm 3 chart that extends the upstream Sonatype Nexus chart found [here](https://github.com/oteemo/charts)]. This deploys Nexus along with the [Nexus Kubernetes plugin](https://github.com/sonatype-nexus-community/nexus-kubernetes-openshift)
--- a/nexus/chart/nexus/.gitignore
+++ b/nexus/chart/nexus/.gitignore
@@ -1,0 +1,2 @@
+Chart.lock
+charts/*
--- a/nexus/chart/nexus/.helmignore
+++ b/nexus/chart/nexus/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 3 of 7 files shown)


pabrahamsson:
Also, do you have any thoughts on what we can around CI for this Chart? I've seen other projects use [bats](https://github.com/bats-core/bats-core) for their testing of helm charts. We also have #313 for other projects in this repo.

tylerauerbeck:
@pabrahamsson  Let me double check this. I think I tested this on a 3.X cluster just to make sure this worked. 

As for testing, I’ve got a few ideas pulled from other repos at this point. Trying to figure out the least amount of overhead here.. so maybe I’ll try to bring a few ideas to the call on Friday so we can figure out how to proceed.

tylerauerbeck:
@pabrahamsson Found the issue with the hostname. Looks like there's a bit of an issue between how `helm template` was handling an empty string versus how `helm install` handles it. This should be fixed and I tested it out on both a 3.X and 4.X cluster.

As for tests, are we okay to merge this before we figure out the testing strategy? Or is this something we'd want to get resolved first?

pabrahamsson:
Thank you @tylerauerbeck, I can verify that it now works on 3.11.

I'm a bit concerned by the following message that is displayed when logging into Nexus.
```
This version of NXRM is vulnerable to a critical remote code execution security vulnerability.
We recommend immediate upgrade to 3.21.2 or later, especially if this is a public instance.
```
Would it be possible to use `3.21.2` or newer image?

tylerauerbeck:
@pabrahamsson Done. Bumped to v3.22.1.

PR #342, opened 2020-05-12
Add Helm chart for RabbitMQ

pabrahamsson:
#### What is this PR About?
A Helm chart for deploying RabbitMQ

#### How do we test this?
Follow instructions in README

### DIFF
--- a/rabbitmq/README.md
+++ b/rabbitmq/README.md
@@ -43,6 +43,8 @@ This is the equivivalent of docker build --build-arg ERLANG_VERSION=19.3.6` to a
 3. `oc new-project rabbitmq`
 4. `helm install rabbitmq chart`
 
+**_NOTE:_** This image is currently not compatible with https://github.com/bitnami/charts/tree/master/bitnami/rabbitmq but this might change in the future.
+
 ### Build and deploy using Openshift Applier
 1. Clone this repository:
    `git clone https://github.com/redhat-cop/containers-quickstarts`
--- a/rabbitmq/README.md
+++ b/rabbitmq/README.md
@@ -34,7 +34,16 @@ The openshift-applier will create the following OpenShift objects:
 `ERLANG_VERSION`& `RABBITMQ_VERSION` are passed on to the buildconfig thus these versions can be controlled in the build.
 This is the equivivalent of docker build --build-arg ERLANG_VERSION=19.3.6` to a docker build.
 
-## Start build and deploy
+## Deploying
+
+### Helm chart
+1. Clone this repository:
+   `git clone https://github.com/redhat-cop/containers-quickstarts`
+2. `cd containers-quickstarts/rabbitmq`
+3. `oc new-project rabbitmq`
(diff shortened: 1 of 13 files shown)


garethahealy:
@pabrahamsson ; any reason not to point users to:
- https://github.com/bitnami/charts/tree/master/bitnami/rabbitmq

does this chart do anything extra than the bitnami one?

pabrahamsson:
> @pabrahamsson ; any reason not to point users to:
> 
> * https://github.com/bitnami/charts/tree/master/bitnami/rabbitmq
> 
> does this chart do anything extra than the bitnami one?

The bitnami chart is [making certain assumptions](https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/statefulset.yaml#L84) about the image that will not work with our image.
```
oc logs rabbitmq-0
mkdir: cannot create directory ‘/opt/bitnami/rabbitmq/.rabbitmq/’: Permission denied
```
@garethahealy do you see a way we can work around this? There are numerous values that can be set in the chart but the above doesn't seem to be configurable.

garethahealy:
that could probably fixed by including the "fix-permissions" script in their image:
- https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/values.yaml#L15-L16

presume you are getting a permissions issue because that folder is owned by root, and the scc is forcing you to run as random 100000000 user

tylerauerbeck:
@garethahealy @pabrahamsson So this is a common thing that we'll run into (and that I have run into) when working with upstream Helm charts. This is especially true for some of the bitnami ones as they've become the "upstream"/ trusted charts for a number of projects.

I think we've talked a bit about this on our weekly calls. But just to restate it here.. I think our goal is to work with upstream as much as we can. But there are definitely going to be times where some work is involved to get there. In the interim, I think it's fine that we carry our own "custom" chart that either augments the upstream chart (by adding to it) or just carry our own chart that works in the mean time while we attempt to work with the upstream chart. Then whenever we do get our changes in place to be able to work with upstream, we can just have our repo provided a functional values file and have instructions that show how to deploy via the upstream chart.

Happy to chat more about this though if there are concerns/questions.

pabrahamsson:
I think there's more investment required to be able to use the bitnami chart that I don't have the time for at this point. If we think this is good enough for now I'd suggest we keep this chart for now. If someone down the road is interested and has the time I'm definitely open to piggyback on the bitnami chart.
@garethahealy @tylerauerbeck thoughts?

tylerauerbeck:
@pabrahamsson That was my thinking

garethahealy:
@pabrahamsson ; sounds fine. maybe just add a comment in the readme, to explain why we didn't use bitnami.

/lgtm

pabrahamsson:
> @pabrahamsson ; sounds fine. maybe just add a comment in the readme, to explain why we didn't use bitnami.
> 
> /lgtm

Sure, I've updated the README

PR #344, opened 2020-05-15
Updated apiVersions to fully qualified 4.x compatible

garethahealy:
#### What is this PR About?
As discussed in CoP meeting. Updates apiVersions to work on 4.x

#### How do we test this?
Apply the template or run rego tests

### DIFF
--- a/.openshift/templates/jenkins-slave-generic-template.j2
+++ b/.openshift/templates/jenkins-slave-generic-template.j2
@@ -1,6 +1,6 @@
 ---
 kind: Template
-apiVersion: v1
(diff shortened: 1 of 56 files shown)


.openshift/templates/jenkins-slave-generic-template.j2
```diff
@@ -1,6 +1,6 @@
 ---
 kind: Template
-apiVersion: v1
+apiVersion: template.openshift.io/v1
```
pabrahamsson, line 3: Since other changes in this PR is reorganizing `apiVersion` to come before `kind` let's do the same here.

PR #345, opened 2020-05-15
Feature/jenkins slave python build

itewk:
#### What is this PR About?
Creates a GitHub action for building the jenkins python worker image.

I did take liberty of calling the image "jenkins-worker-python" vs "jenkins-slave-python". Figure we can rename the folders later but figure if we are going to publish these images we should use the better name.

#### How do we test this?
I have tested this on my fork https://github.com/itewk/containers-quickstarts and my repo quay.io/itewk


garethahealy:
@itewk ; if the name is due to the history of the word "slave", I'd suggest using "agent" instead of "worker", as its commonly accepted as the replacement name.

itewk:
@garethahealy updated the mere request. Forgot to mention but before this gets merged someone with admin to quay.io/redhat-cop needs to go in and create `jenkins-agent-python` and give the robot account GitHub is configured to use `write` access to the repo.

PR #349, opened 2020-05-20
Add Bats ubi8 utility image

pabrahamsson:
#### What is this PR About?
Adds a utility image containing [Bats](/bats-core/bats-core) and also
- Helm
- jq
- Openshift Client
- yq

The image can be used standalone to run bats test scenarios
```
podman run --rm -v $PWD:/code ubi8-bats /code/unit
```
It can also be used as part of a CI pipeline.

>**NOTE:** The required Quay repo has already been created: https://quay.io/repository/redhat-cop/ubi8-bats

#### How do we test this?
See included GH workflows


.github/workflows/ubi8-bats-pr.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Build image
+        uses: docker/build-push-action@v1
+        with:
+          path: ${{ env.context }}
+          push: false
+          repository: ${{ env.image_name }}
+          tags: ${{ steps.check_version.outputs.IMAGE_TAGS }}
+      - name: Test image
+        run: |
+          echo "Running: docker run --entrypoint bash ${image_name}:${{ steps.check_version.outputs.IMAGE_TAGS }} -c 'bats --version'"
```
garethahealy, line 28: @pabrahamsson ; can we not use bats to do all these tests?

utilities/ubi8-bats/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
garethahealy, line 1: should we start using the .io registry? needs auth, not sure if that's been thought about when doing a build via gh-actions

.github/workflows/ubi8-bats-pr.yaml
pabrahamsson: We sure can, I've updated the PR

utilities/ubi8-bats/Dockerfile
pabrahamsson: Are there benefits using `registry.redhat.io` over `registry.access.redhat.com`?

JaredBurck: `registry.access.redhat.com` will be decommissioned sometime in the coming year. It is recommended that we prepare to use `registry.redhat.io`.  However, that registry does require authentication and supports registry tokens.

https://access.redhat.com/RegistryAuthentication

pabrahamsson: Thanks @JaredBurck I think I knew that was coming but didn't have a timeline.
It looks as though the Docker GH action that we use supports auth for pulling and thus it should be possible to use `registry.redhat.io`.
@etsauer @sabre1041 thoughts on changing registry?

garethahealy:
merging. can raise another PR if we decide to move base images to io registry.

PR #351, opened 2020-05-23
Python Kopf S2I examples and docs

jkupferer:
#### What is this PR About?

This PR adds a simpler example for how to use the Python Kopf s2i image as well as documentation of how to use this s2i image with `odo` for in-cluster operator development.

#### How do we test this?

Documentation included in this PR covers testing instructions. Update to GitHub workflows should also process automatically to verify examples still function to test the build.


jkupferer:
Marking this as WIP, just noticed that it is a bit soon to assume everyone has `oc` with `kustomize` support.

jkupferer:
Updated the examples and documentation to not require kustomize.

PR #352, opened 2020-05-26
upload-file-to-google-drive.py - init add

itewk:
#### What is this PR About?
adds upload-file-to-google-drive.py to ubi8-google-api-python-client image.

#### How do we test this?
Updated github PR workflow to have a basic test that the script is there.

### DIFF
--- a/utilities/ubi8-google-api-python-client/README.md
+++ b/utilities/ubi8-google-api-python-client/README.md
@@ -8,6 +8,76 @@ Useful for environments where these libraries/tools are needed such as:
 * local development
 * CI tools that do not require specific CI tooling configuraiton, such as GiLab Jobs.
 
+## Helper Scripts
+
+This image also contains custom helper scripts for interacting with the Google API.
+
+### upload-file-to-google-drive.py
+This script makes an easy way for uploading files to Google Drive into existing or new folders.
+
+#### Parameters
+
+| parameter                | required | default | comments
+|--------------------------|----------|---------|---------
+| --credentials            | yes      |         | A JSON file with google API crednetials.
+| --drive-id               | no       |         | If wanting to upload to a Google Team Share drive then this is where to specivy that share drive id.
+| --file                   | yes      |         | Local file path of the file to upload.
+| --parent-drive-folder-id | yes      |         | Google Drive Folder ID to be the parent of the file (or sub folder containing the file if sub-folder-name supplied).
+| --sub-folder-name        | no       |         | Name of a sub folder to create (if it doesn't exist) under the parent-drive-folder-id folder to then put the uploaded file in.
+| --destination            | yes      |         | Name of the file when uploaded to Google Drive.
+| --dest-mime-type         | no       | `application/vnd.google-apps.document`                                    | MIME type of the destination file once uploaded, this can be used to have google automatically convert a file from one type to another, say from DOCX to gDoc.
+| --source-mime-type       | no       | `application/vnd.openxmlformats-officedocument.wordprocessingml.document` | MIME type of the source file.
+
+#### Getting a credentials.json
+There are many ways to do this. The way in which the author of this script did it was to follow steps from
+[How to use GCP service accounts with Google Apps Script projects to automate actions in G Suite](How to use GCP service accounts with Google Apps Script projects to automate actions in G Suite) specfiically "SECTION TWO: Create a GCP project, a service account, activate the Google Drive API, and an API key". No matter the instructions you follow
+ultimatlly what you need to end up looks like the following.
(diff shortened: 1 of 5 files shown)


utilities/ubi8-google-api-python-client/scripts/upload-file-to-google-drive.py
```diff
@@ -0,0 +1,263 @@
+from __future__ import print_function
```
jkupferer, line 1: Since this is meant to be run with python3, it should start with:
```
#!/usr/bin/env python3
```

utilities/ubi8-google-api-python-client/README.md
```diff
@@ -8,6 +8,39 @@ Useful for environments where these libraries/tools are needed such as:
...
+  --sub-folder-name="${GOOGLE_DRIVE_SUB_FOLDER_NAME_UNDER_PARENT_DRIVE_FOLDER_ID_TO_UPLOAD_FILE_TO}" \
+  --destination="${GOOGLE_DRIVE_FILE_NAME}" \
+  --dest-mime-type='${DESTINATION_FILE_TYPE}' \
+  --source-mime-type='${SOURCE_MIME_TYPE}' \
+  --credentials="${google_api_credentials_json}"
+```
+
+Upload DOCX and convert to GDOC
+```bash
+python3 /scripts/upload-file-to-google-drive.py \
```
jkupferer, line 33: `upload-file-to-google-drive.py` should be executable with a "shebang", `#!/usr/bin/env python3`, line so that python3 does not need to be explicitly referenced.

Also, why not put this script in a standard location such as `/usr/local/bin`?

utilities/ubi8-google-api-python-client/README.md
```diff
@@ -8,6 +8,39 @@ Useful for environments where these libraries/tools are needed such as:
...
+  --file=foo.docx \
+  --drive-id="${GOOGLE_DRIVE_ID}" \
```
garethahealy, line 35: can you add a section to the readme as to how I'd get these envs?

utilities/ubi8-google-api-python-client/scripts/upload-file-to-google-drive.py
```diff
@@ -0,0 +1,263 @@
...
+import sys, getopt
```
garethahealy, line 2: just wondering if there is a better place for this script. can it be used by other things? could it live it a repo of python helper scripts?

utilities/ubi8-google-api-python-client/scripts/upload-file-to-google-drive.py
itewk: @jkupferer i didn't put that in since some systems don't have `python3` they just have `python` which happens to be `python3` I didn't want to hard code in any script paths and thought it better to just have people run it using whichever python library they had wherever they had it. not to mention the complications when using virtual environments.

utilities/ubi8-google-api-python-client/README.md
itewk: Also, why not put this script in a standard location such as /usr/local/bin

> because it itself is not executable.

utilities/ubi8-google-api-python-client/scripts/upload-file-to-google-drive.py
itewk: in theory, but long story short, it was living in another internal repo where we didn't want it any longer. that repo uses this image. this script requires having all the google python libraries available. it seems easiest just to keep it with the container. if someone else somewhere down the road wants to use it outside the container they could do the work to move it. but the juice doesn't seem worth the squeeze to put it in some other repo and then import it during image build.

itewk:
i have not forgotton about this, will do all the requested updates. Life.

utilities/ubi8-google-api-python-client/scripts/upload-file-to-google-drive.py
itewk: updated to to do this in latest push.

utilities/ubi8-google-api-python-client/README.md
itewk: done.

utilities/ubi8-google-api-python-client/README.md
itewk: updated to do this.

itewk:
@pabrahamsson updated MR and resolved all conversations.

itewk:
@pabrahamsson thanks for the merge.
@garethahealy and @jkupferer thank you for the review comments.

PR #357, opened 2020-06-03
🎣 Cheeky bump to the node version, removal of chrome and base ubi  🤹‍♀️

springdo:
#### What is this PR About?
Cheeky bump to the node version. 

#### How do we test this?
You don't - Prow will do it for us :) 


springdo:
You love these tiny changes :P 
/assign @pabrahamsson

garethahealy:
@springdo ; btw, if you do a rebase (https://thoughtbot.com/blog/keeping-a-github-fork-updated) you don't get all the commit history that's not related to this PR

springdo:
@garethahealy  - yeah i was being lazy and I know you'd squash the commits anyways so

springdo:
Yeah I can update that @pabrahamsson  - but just a thought, we're using a rhel dockerfile too in this repo and there is no rhoar-node12 afaik. I can look at bringing both together again and probs using a more upto date slave base?

pabrahamsson:
> Yeah I can update that @pabrahamsson - but just a thought, we're using a rhel dockerfile too in this repo and there is no rhoar-node12 afaik. I can look at bringing both together again and probs using a more upto date slave base?

That would be great @springdo 

As you might have seen for most of the other Jenkins Agents in this repo they've been consolidated to use the UBI based `quay.io/openshift/origin-jenkins-agent-base:4.x` as their base. The benefit here has been that we no longer need to manage multiple Dockerfiles.

IIRC what complicates things for this specific agent is the inclusion of the Chrome browser which has dependencies not found in the UBI yum/dnf repositories.

I also think you're probably the most familiar with this particular agent so curious to hear your thoughts here.

springdo:
Yup - I was just messing about with that and discovered all those deps for Chrome are not around. 

I guess we could look at a few things - I use Chrome in this image for doing headless testing but some of the test frameworks now package and download Chrome (or Chromium) too. Recently saw an example using this for their Karma tests, but not their protractor tests. 
https://www.npmjs.com/package/puppeteer 

I'll see if it works as an alternative though then having it packaged in the salve, thus making this one easier to maintain. Of course downloading a browser on every CI run has it's own draw backs. 

Given we only have 3 deps, perhaps we could add non subscription based repos for those extra chrome deps? Although before i go down that rabbit hole there are probs countless dependencies for those dependencies..... hmmm

garethahealy:
fyi: Chromium is in epel.

PR #373, opened 2020-06-11
Fix kopf peering config and bump version

jkupferer:
#### What is this PR About?

This PR moves to the latest release of kopf and fixes a bug in the KOPF_PEERING option.

#### How do we test this?

Use tests and described in the build-s2i-python-kopf dir


build-s2i-python-kopf/README.adoc
```diff
@@ -7,6 +7,15 @@ Source-to-Image for python https://kopf.readthedocs.io/en/latest/[kopf] based op
 This Python Kopf operator source-to-image can be used to build python-base operators using OpenShift source-to-image or Dockerfile based builds.
 This s2i image may also be used with `odo` to do in-cluster operator development.
 
+=== Kopf Peering CustomResourceDefinitions
+
+Unless Kopf standalone mode is used, peering between Kopf operators should be enabled.
+The Kopf peering custom resource definitions are required for peering and can be installed with:
+
+--------------------------------------------------------------------------------
+oc create -f https://raw.githubusercontent.com/redhat-cop/containers-quickstarts/master/build-s2i-python-kopf/peering-crds.yaml
```
sabre1041, line 16: Can `create` be replaced with `apply`?

build-s2i-python-kopf/version.json
```diff
@@ -1 +1 @@
-{"version":"v0.28.2"}
+{"version":"v0.27.0"}
```
sabre1041, line 1: Version decrement?

build-s2i-python-kopf/README.adoc
jkupferer: I think `apply` would be okay here.

build-s2i-python-kopf/version.json
jkupferer: Yeah... this is a weird case. The version wasn't supposed to be 0.28 to begin with. The plan was to match the version to the python kopf version, which is now at 0.27. I'm not sure how I ended up creating the image version at 0.28. I'm torn about whether I should continue at 0.28 until kopf catches up, or go ahead and just match the version immediately like this. What do you think?

jkupferer: I suppose waiting for kopf to catch up is probably the more obvious choice and less likely to create mystery...

PR #376, opened 2020-06-18
🦩bump version to latest argocd-1.6.0🦩

eformat:
#### What is this PR About?

#### How do we test this?

### DIFF
--- a/jenkins-slaves/jenkins-slave-argocd/Dockerfile
+++ b/jenkins-slaves/jenkins-slave-argocd/Dockerfile
@@ -1,7 +1,7 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:4.4
 
-ENV ARGOCD_VERSION=1.5.6 \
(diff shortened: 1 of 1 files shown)


garethahealy:
@pabrahamsson ; what causes the _needs-ok-to-test_ label to be added? I would have thought if a change affects the Jenkins slaves -  which has the prow job attached - the tests would auto-trigger, but that didn't happen.

pabrahamsson:
> @pabrahamsson ; what causes the _needs-ok-to-test_ label to be added? I would have thought if a change affects the Jenkins slaves - which has the prow job attached - the tests would auto-trigger, but that didn't happen.

I believe sure this happens when the committer isn't part of the `redhat-cop` org. 

So @eformat you might want to open an issue over at [redhat-cop/org](/redhat-cop/org/issues/new/choose) to become a new contributor. This is obviously not a requirement but more of a convenience.

PR #379, opened 2020-06-19
Fix typos and update example in README

bparry02:
Update the example commands to genericize the use case.
Add command to add gitlab helm repo

#### What is this PR About?
Updating the readme to clarify commands

#### How do we test this?
n/a


bparry02:
@itewk FYI

PR #380, opened 2020-06-22
increasing go version in jenkins-slave-golang to 1.14.4 and adding tests

deweya:
#### What is this PR About?

* Increased the go version of jenkins-slave-golang to `1.14.4`. Upgrading to Go >1.11 will let users leverage Go modules.

* Removed `dep` from this image since newer versions of Go rely on modules. Dep recognizes this also on [their repo](https://github.com/golang/dep#dep):

    > dep was initially developed to experiment with a new dependency management system for Go; but, as of Go 1.11, the Go project has officially adopted a different approach, based on the concept of Modules

* Added a few things to help test this and future changes. I added [this hello world app](https://github.com/go-modules-by-example/index/tree/master/001_go_modules_tour) which can be built and executed in Jenkins to ensure Go apps can be built properly. I also added to Jenkinsfile.test so that the test can be easily run.

* To allow tests to run successfully, I gave `/usr/src/go` root group-writable permission so that users don't need to make any modifications on their end to build their app. I also had to remove this line from the Dockerfile:

    > chown -R 1001 ${HOME}/.cache/go-build

    because it returned "No such file or directory".

#### How do we test this?

After building jenkins-agent-golang in OCP (I tested in 4.4), run Jenkinsfile.test in a multibranch pipeline and make sure that:
* The Go version prints
* The app can be built
* The app is able to run

### DIFF
--- a/jenkins-slaves/jenkins-slave-golang/Dockerfile
+++ b/jenkins-slaves/jenkins-slave-golang/Dockerfile
@@ -1,6 +1,6 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:4.4
 
-ENV GO_VERSION_DEFAULT=1.10.2 \
+ENV GO_VERSION_DEFAULT=1.14.4 \
     GOROOT=/usr/local/go \
     GOPATH=/usr/src/go
 ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
@@ -16,10 +16,9 @@ RUN if [ -z $GO_VERSION ] ; then GO_VERSION=${GO_VERSION_DEFAULT} ; fi && \
     mkdir -p /usr/src/go/src/redhat && \
     tar -xzf /usr/go${GO_VERSION}.linux-amd64.tar.gz && \
     mv $(pwd)/go /usr/local/ && \
-    go get -u github.com/golang/dep/cmd/dep && \
     chown -R 1001 /usr/src/go && \
     chown -R 1001 /usr/local/go && \
-    chown -R 1001 ${HOME}/.cache/go-build && \
+    chmod g+w /usr/src/go && \
     rm -f /usr/go${GO_VERSION}.linux-amd64.tar.gz
 
 USER 1001
--- a/jenkins-slaves/jenkins-slave-golang/README.md
+++ b/jenkins-slaves/jenkins-slave-golang/README.md
@@ -17,4 +17,7 @@ oc process -f ../../.openshift/templates/jenkins-slave-generic-template.yml \
 For all params see the list in the `../../.openshift/templates/jenkins-slave-generic-template.yml` or run `oc process --parameters -f ../../.openshift/templates/jenkins-slave-generic-template.yml`.
 
 ## Jenkins
-Add a new Kubernetes Container template called `jenkins-slave-golang` (if you've build and pushed the container image locally) and specify this as the node when running builds. If you're using the template attached; the `role: jenkins-slave` is attached and Jenkins should automatically discover the slave for you. Further instructions can be found [here](https://docs.openshift.com/container-platform/3.7/using_images/other_images/jenkins.html#using-the-jenkins-kubernetes-plug-in-to-run-jobs).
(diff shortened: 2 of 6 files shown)


jenkins-slaves/jenkins-slave-golang/README.md
```diff
@@ -17,11 +17,4 @@ oc process -f ../../.openshift/templates/jenkins-slave-generic-template.yml \
...
-```
-export GOPATH=${WORKSPACE}
-go get -v -t ./...
-go build -v
-# if there are Ginkgo tests then also run this!
-$GOPATH/bin/ginkgo -r --cover -keepGoing
-```
+Add a new Kubernetes Container template called `jenkins-slave-golang` (if you've build and pushed the container image locally) and specify this as the node when running builds. If you're using the template attached; the `role: jenkins-slave` is attached and Jenkins should automatically discover the slave for you. Further instructions can be found [here](https://docs.openshift.com/container-platform/3.7/using_images/other_images/jenkins.html#using-the-jenkins-kubernetes-plug-in-to-run-jobs).
```
pabrahamsson, line 20: Can you add something to the README about go dep having been removed and now using go modules?

deweya: Sure. Added a short Modules section

jenkins-slaves/jenkins-slave-golang/README.md
```diff
@@ -17,11 +17,7 @@ oc process -f ../../.openshift/templates/jenkins-slave-generic-template.yml \
...
```
garethahealy, line 20: can you update the docs link to be 4.x

jenkins-slaves/jenkins-slave-golang/Dockerfile
```diff
@@ -16,10 +16,9 @@ RUN if [ -z $GO_VERSION ] ; then GO_VERSION=${GO_VERSION_DEFAULT} ; fi && \
     mkdir -p /usr/src/go/src/redhat && \
     tar -xzf /usr/go${GO_VERSION}.linux-amd64.tar.gz && \
     mv $(pwd)/go /usr/local/ && \
-    go get -u github.com/golang/dep/cmd/dep && \
     chown -R 1001 /usr/src/go && \
     chown -R 1001 /usr/local/go && \
-    chown -R 1001 ${HOME}/.cache/go-build && \
+    chmod g+w /usr/src/go && \
```
garethahealy, line 21: whats `/usr/src/go` used for?

jenkins-slaves/jenkins-slave-golang/go.mod
```diff
@@ -0,0 +1,8 @@
+module example.com/hello
```
garethahealy, line 1: since go bits added are for a test app. should they live under an "example" or "test" dir? just thinking not to confuse people as to why there are some go files in the root dir.

@pabrahamsson ; thoughts?

jenkins-slaves/jenkins-slave-golang/main.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright (c) 2018 Paul Jolly <paul@myitcv.io>. All rights reserved.
```
garethahealy, line 2: is it possible to provide a different example, so it's not copyrighted?

@pabrahamsson ; thoughts?

deweya: I chose that example in order to make sure go modules were being handled properly

jenkins-slaves/jenkins-slave-golang/Dockerfile
deweya: It can be, for sure. This was the path used before this PR, but typically GOPATH is `${HOME}/go`

jenkins-slaves/jenkins-slave-golang/go.mod
deweya: I agree with you. Should I go ahead and make this change @pabrahamsson ?

jenkins-slaves/jenkins-slave-golang/README.md
deweya: Updated

PR #382, opened 2020-06-23
Update ruby jenkins agent oc client to 4.4

tom-stockwell:
#### What is this PR About?
Updating the ruby jenkins agent oc client to 4.4.

Alternatively, we could remove the custom installation of the `oc` client altogether and depend entirely on the base image to provide it. However, it does provide a slightly older version of the client as demonstrated below.

```console
$ podman run --rm --entrypoint "/bin/bash" quay.io/openshift/origin-jenkins-agent-base:4.4 -c 'oc version'     
Client Version: v4.2.0-alpha.0-459-gd038424
```

Does anyone have any thoughts regarding this?

#### How do we test this?
Standard ruby agent tests.

### DIFF
--- a/jenkins-slaves/jenkins-slave-ruby/Dockerfile
+++ b/jenkins-slaves/jenkins-slave-ruby/Dockerfile
@@ -14,7 +14,7 @@ It is simple, straight-forward, and extensible." \
   PATH=$PATH:/home/jenkins/bin \
   PROMPT_COMMAND=". /opt/app-root/etc/scl_enable" \
   HOME=/home/jenkins \
-  ORIGIN_CLIENT=https://mirror.openshift.com/pub/openshift-v3/clients/3.11.50/linux/oc.tar.gz
+  OC_VERSION=4.4
 
 
 LABEL summary="$SUMMARY" \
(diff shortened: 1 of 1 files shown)


jenkins-slaves/jenkins-slave-ruby/Dockerfile
```diff
@@ -14,7 +14,7 @@ It is simple, straight-forward, and extensible." \
   PATH=$PATH:/home/jenkins/bin \
   PROMPT_COMMAND=". /opt/app-root/etc/scl_enable" \
   HOME=/home/jenkins \
-  ORIGIN_CLIENT=https://mirror.openshift.com/pub/openshift-v3/clients/3.11.50/linux/oc.tar.gz
+  ORIGIN_CLIENT=https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.4/linux/oc.tar.gz
```
garethahealy, line 17: @tom-stockwell ; can you change the version to be an env? i.e.: OC_VERSION

this is how 99% of the other images do it, just to keep things consistent.

garethahealy:
CI failure is a flake not related to this PR.

jenkins-slaves/jenkins-slave-ruby/Dockerfile
tom-stockwell: Sure

tom-stockwell: I noticed that `utilities/ubi8-bats/Dockerfile` uses an `ARG` while `tool-box/Dockerfile` uses an `ENV`. Any preference one way or the other?

garethahealy: don't mind, the build process which releases these images doesn't pass in the value as an arg, so there isn't any benefit either way

Issue #395, opened 2020-07-29
Add Opinionated Helm 3 Chart For Jenkins On OpenShift


InfoSec812:
Add a Helm 3 Chart for Jenkins on OpenShift

PR #396, opened 2020-07-29
Added Helm 3 chart for Jenkins on OpenShift

InfoSec812:
#### What is this PR About?
Adds a Helm 3 Chart for deploying Jenkins on OpenShift

Resolves: #395 

#### How do we test this?
```
git clone <repo>
cd <repo>/jenkins-helm3
helm template jenkins ${PWD}
```

### DIFF
--- a/jenkins-helm3/.helmignore
+++ b/jenkins-helm3/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/jenkins-helm3/Chart.yaml
+++ b/jenkins-helm3/Chart.yaml
@@ -1,0 +1,21 @@
+apiVersion: v2
+name: Jenkins
+description: A Helm chart for OpenShift
+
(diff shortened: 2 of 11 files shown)


tylerauerbeck:
@InfoSec812 We're currently working on moving the rht-labs/charts repo into the CoP, which has a Jenkins chart as well. We should take a look to see if there is any delta between the two and then get that merged in there. I'll check to see where we are in the migration process and then see if we can get this moved there if that's alright?

InfoSec812:
@tylerauerbeck Looking at the Labs Chart for Jenkins, I think that this is pretty specific to the Labs way of doing things. I think we'll want to use some logic to allow different installation paths/configs/options... For example, SealedSecrets...

springdo:
Hi @InfoSec812 that sounds great but perhaps there is a way we can add those config specific versions in such a way that both the labs and cop approaches are maintained from one single source. As we all know, there are about 8 different ways across cop repos to install Jenkins right now from templates .... Would love to not be in same position with the helm approaches!

BTW, Jenkins in the labs space does support sealed secrets at the moment if you want to try it out!

InfoSec812:
@springdo My point in the comment is that I **CANNOT** use SealedSecrets in the deployment environments I am targeting right now. Which is why the Jenkins Chart needs to be a little *less* opinionated (or at least more flexible).

springdo:
Hey Deven! The rht-labs helm-charts repository has moved across to the cop-org, do you want some help to create a values file to match your implementation here or how do you want to proceed?

pabrahamsson:
@InfoSec812 @tylerauerbeck @springdo can we try and figure out where we want to go with this? I agree that having multiple Helm charts for Jenkins isn't ideal, will this [Jenkins Helm chart](https://github.com/redhat-cop/helm-charts/tree/master/charts/jenkins) work for your use case @InfoSec812?

PR #401, opened 2020-08-13
Jenkins agent for python 3.8

mabulgu:
#### What is this PR About?
This PR includes the Jenkins agent for python 3.8

#### How do we test this?
Please see the README.md of the agent for more details.


garethahealy:
@mabulgu ; any reason not to just bump the python agent to 3.8? I presume 3.8 is backwards compatible with 3.6?

mabulgu:
> @mabulgu ; any reason not to just bump the python agent to 3.8? I presume 3.8 is backwards compatible with 3.6?

Hi @garethahealy. Using the python 3.6 may cause problems esp. with frameworks like Django which has many dependant libraries like SQLite in its spine. For example with the default db configuration that comes with Django's latest version one can not run any tests or do stuff with manage.py because it will complain about the current version compatibility of sqllite which will cause downgrading Django 3.x to Django 2.x which is a huge version change.

In case people may not want to run an external db in the first place and use SQLite for this example (or just run the tests with SQLite is a common approach too), I thought it would be a good idea to create an agent for python 3.8.

This was actually a recent problem we faced in a recent project: we are running Django 3.x with python 3.8 base image but, for the build stuff we needed this slave for 3.8 regarding the issues I mentioned above.

mabulgu:
Oh, I think I misunderstood you @garethahealy, sorry for that. Yeah, we may just bump the version up to 3.8 if it is ok for everybody. It will work for all 3.6 apps too.

pabrahamsson:
> Oh, I think I misunderstood you @garethahealy, sorry for that. Yeah, we may just bump the version up to 3.8 if it is ok for everybody. It will work for all 3.6 apps too. I wasn't just sure about it may break some apps and maybe people may want to use the previous stable version..

:+1: for bumping the existing python agent to 3.8. If people still want to use 3.6 they can use the current latest release of this repo.

mabulgu:
Perfect! You are 100% right, I will change it in this current PR and let you know. Thanks

pabrahamsson:
@mabulgu have you had a chance to look into updating the existing python agent?

mabulgu:
Hi. I will change the agent for python and update it shortly. I used the one (3.8) for a labs engagement but could not have time to bump the current version. Will do it shortly. Thanks for the ping👋

mabulgu:
Sorry for it to take so long; had many things in the queue:/ I checked that agent image is created successfully. @pabrahamsson

pabrahamsson (approved):
Thank you @mabulgu, looks good and tests are passing 
/lgtm

PR #403, opened 2020-09-03
Cert Management: Adding OCP v4 templates for use with OpenShift Applier

oybed:
#### What is this PR About?
Adding support for cert-manager / OCPv4 for use with openshift-applier

#### How do we test this?
Tweak the params, then run the openshift-applier with the inventory

### DIFF
--- a/cert-manager-configs/.applier/group_vars/seed-hosts.yml
+++ b/cert-manager-configs/.applier/group_vars/seed-hosts.yml
@@ -1,0 +1,32 @@
+---
+
+openshift_domain: "my.cluster.example.com"
+cert_manager_namespace: "cert-manager"
+
+cert_manager_params:
+  DNS_DOMAIN: "{{ openshift_domain }}"
+  AWS_ACCESS_KEY_UD: "ASDF321ADF63"
+  AWS_SECRET_ACCESS_KEY: "$#%DASFSRT@#$DFFSVSDFG@#$@#$"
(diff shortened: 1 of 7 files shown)


cert-manager-configs/.applier/group_vars/seed-hosts.yml
```diff
@@ -0,0 +1,32 @@
+---
+
+openshift_domain: "my.cluster.example.com"
+cert_manager_namespace: "cert-manager"
+
+cert_manager_params:
+  DNS_DOMAIN: "{{ openshift_domain }}"
+  AWS_ACCESS_KEY_UD: "ASDF321ADF63"
```
pabrahamsson, line 8: Typo? Should be `AWS_ACCESS_KEY_ID` I assume.

pabrahamsson (changes requested):

cert-manager-configs/.applier/group_vars/seed-hosts.yml
oybed: yes, fixed.

pabrahamsson (commented):
Looks good @oybed, should we make a mention somewhere that the `AWS_SECRET_ACCESS_KEY` needs to be base64 encoded? Alternatively we could use `stringData` in the secret.

pabrahamsson (approved):
Thank you @oybed 
/lgtm

PR #410, opened 2020-09-21
Make utilities/ubi8-asciidoctor image significantly smaller

nesanton:
#### What is this PR About?
* Use ubi8 instead of ubi8-minimal.
  When installing so much (esp. python) there's barely any difference.
  Yet, one can leverage full dnf in ubi8
* Less layers with curl | tar
* Install gcc, make etc. -> install gems -> remove gcc, make etc. all in one layer
  This saves up to 300M
* Cleanup cache and docs where reasonable

#### How do we test this?

Create the image and check the output of the pdf ....
```bash
podman run --rm --name asciidoctor -v //:/documents/ -w /documents/styles quay.io/redhat-cop/ubi8-asciidoctor:1.3 asciidoctor-pdf -r asciidoctor-diagram -a ej-base-dir=/documents/<document.adoc>
```

### DIFF
--- a/utilities/ubi8-asciidoctor/Dockerfile
+++ b/utilities/ubi8-asciidoctor/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8-minimal
+FROM registry.access.redhat.com/ubi8
 
 LABEL MAINTAINERS="Red Hat Services"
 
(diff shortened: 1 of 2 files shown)


nesanton:
@buuhsmead, I guess your blessing can be useful too.

I built it to quay.io/anestero/ubi8-asciidoctor-slim for testing

buuhsmead:
@nesanton This merge quay.io/anestero/ubi8-asciidoctor-slim gives the same output as v1.2. So the same functional output.

buuhsmead:
From 806MB down to 546MB image size.

nesanton:
> @nesanton This merge quay.io/anestero/ubi8-asciidoctor-slim gives the same output as v1.2. So the same functional output.

yeah, I also tested it on a few of my adocs  for pdf and html. The results are identical

utilities/ubi8-asciidoctor/Dockerfile
```diff
@@ -64,7 +66,9 @@ RUN gem install --no-document \
     rouge \
     slim \
     thread_safe \
-    tilt
+    tilt \
+  && rm -rf /usr/local/share/gems/cache \
+  && dnf history undo last -y && dnf clean all && rm -rf /var/lib/dnf
```
pabrahamsson, line 71: TIL `dnf history undu last` :heart_eyes_cat:

pabrahamsson (approved):
Thank you @nesanton
/lgtm

PR #419, opened 2020-11-04
Update Dockerfile

adynetro:
add curl to image 


pabrahamsson (commented):
Thank you @adynetro, can you give some context around why curl is needed in this image?

adynetro:
curl is not present in version greather than 8.1 and installplugin.sh fails @pabrahamsson

pabrahamsson (approved):
/lgtm

PR #420, opened 2020-11-18
fix images to java:8 and nodejs:10 for hygieia

welshstew:
#### What is this PR About?
Fixing image versions (nodejs:8 not in 4.5 openshift and java:latest will not build - so fixed to java:8)
#### How do we test this?
Ensure playbook runs and app stands up.  Tested locally with:
```
export ANSIBLE_INVALID_TASK_ATTRIBUTE_FAILED=False
ansible-playbook -i .applier/ galaxy/openshift-applier/playbooks/openshift-cluster-seed.yml -e filter_tags=all -e skip_version_checks=true
```

### DIFF
--- a/hygieia/.openshift/templates/build-base-image.yml
+++ b/hygieia/.openshift/templates/build-base-image.yml
@@ -29,7 +29,7 @@ objects:
       dockerStrategy:
         from:
           kind: ImageStreamTag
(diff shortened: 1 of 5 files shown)


garethahealy (approved):
bu-ti-ful

PR #429, opened 2020-12-22
🐍 Feature/jenkins agent mvn mandrel 🐍

eformat:
#### What is this PR About?
The mandrel drop in replacement for graal is license permissive for building quarkus images
This is a bit of an uber java image - allowing you to just get on and do jdk/mvn based builds including native-image if you need.
Packages helm, yq, jq, oc because they are all useful

Inspired by this upstream image:
quay.io/quarkus/ubi-quarkus-mandrel:20.3-java11

Likely there is work to do to slim this down, see if we can extend base jenkins, but its useful to have latest versions (esp. mvn) deployed as is. Since i have been using it, i thought it might be useful to share as-is, and work on improving it.

#### How do we test this?
Jenkinsfile.test


garethahealy (changes requested):
can you add a comment to the contrib/go-init files as to where they came from. presume these have been pulled from another repo?

just thinking about if these need updating in the future.

eformat:
@garethahealy - done. yeah, ideally there would be a UBI redistributable/buildable base jenkins image? i'm not aware of one. so for now have taken this approach which i know is not ideal.

pabrahamsson:
Thank you for this @eformat 
I think this should replace the existing [jenkins-agent-graalvm](/redhat-cop/containers-quickstarts/blob/master/jenkins-agents/jenkins-agent-graalvm) rather than adding a new agent.
I also think we can use the upstream [origin-jenkins-agent-base](https://quay.io/openshift/origin-jenkins-agent-base) and just add any additional tools (mandrel, maven, jq, yq, helm, ....). The upstream images are built the same/similar way as your Dockerfile.
Happy to merge as is though, we can iterate on the build details.

eformat:
closing in favour of - https://github.com/redhat-cop/containers-quickstarts/pull/437

PR #435, opened 2021-02-05
Update cert-utils-operator target namespaces

paulbarfuss:
#### What is this PR About?

Latest cert-utils-operator supports `AllNamespaces` rather than `SingleNamespace` install mode. Adding an empty list will allow the operator installation to properly complete.

#### How do we test this?
This can be installed independently with the following

```bash
oc process --local -f cert-manager-configs/.openshift/cert-utils-operator.yml | oc apply -f -
```


oybed (approved):
LGTM

PR #436, opened 2021-02-10
🐝 argocd 1.8.3, yq 4.5.0 🐝

eformat:
#### What is this PR About?
bump to latest supported versions

#### How do we test this?
see jenkinsfile test

### DIFF
--- a/jenkins-agents/jenkins-agent-argocd/Dockerfile
+++ b/jenkins-agents/jenkins-agent-argocd/Dockerfile
@@ -1,7 +1,7 @@
 FROM quay.io/openshift/origin-jenkins-agent-base:4.6
 
-ENV ARGOCD_VERSION=1.8.1 \
-    YQ_VERSION=3.4.1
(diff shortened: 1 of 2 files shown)


pabrahamsson (approved):
LGTM @eformat 
Just want to make sure you're aware of some pretty [significant changes](https://mikefarah.gitbook.io/yq/v/v4.x/upgrading-from-v3) in the yq v4 release. 
I'm not familiar with how yq is used in this container but happy to merge as long as you're ok with the above changes.

eformat:
hi @pabrahamsson .. yes, i have felt some pain .. however.. its not so bad once you get the hang of hit

yq was added in at later date, it was not in original here - https://github.com/redhat-cop/containers-quickstarts/commit/466cf26158d6614f97bd430fbefac9a5f60f7bfe

but yeah, upgrading to 4 is breaking change if you use it in combination with argocd cli

PR #448, opened 2021-03-02
asciidoctor-pdf image: Move home out of a mounted volume

nesanton:
#### What is this PR About?
some commands, like `git config --global` are failing due to missing or inappropriate HOME


garethahealy:
I think this is used by the CER project - presume this change doesn't break anything in there?

CC @itewk

itewk:
@garethahealy yeppers. @nesanton is working in the context of the CER project so this is all in line with that.

garethahealy (approved):

PR #492, opened 2021-04-15
Add tower custom buildconfig

paulbarfuss:
#### What is this PR About?
This quickstart gives an example of how to deploy a custom Ansible Tower image with custom virtualenvs and RPMs directly to your Openshift internal image registry to be consumed by the Ansible Tower 3.8 installer.

#### How do we test this?
Following the steps in the README will yield a buildconfig and an imagestream in your Openshift cluster. Run the applier, and validate:

```bash
$ oc get buildconfig tower-ocp -n tower-ocp
$ oc get imagestream tower-ocp -n tower-ocp
```

### DIFF
--- a/_test/conftest.sh
+++ b/_test/conftest.sh
@@ -196,6 +196,17 @@ setup_file() {
   [ "$status" -eq 0 ]
 }
 
+@test "tower-ocp-custom/.openshift" {
+  tmp=$(split_files "tower-ocp-custom/.openshift")
+
+  namespaces=$(get_rego_namespaces "ocp\.deprecated\.*")
+  cmd="conftest test ${tmp} --output tap ${namespaces}"
+  run ${cmd}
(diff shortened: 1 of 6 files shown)


garethahealy (changes requested):
can you add a test to conftes.sh similar to:
- https://github.com/redhat-cop/containers-quickstarts/blob/master/_test/conftest.sh#L23-L32

the key bit, is:
- https://github.com/redhat-cop/containers-quickstarts/blob/master/_test/conftest.sh#L24

which points at the yaml resources (i.e.: your template)

paulbarfuss:
Thanks @garethahealy that was very straightforward and looks to be passign with an `ok` now.

oybed (approved):
LGTM

garethahealy (approved):

PR #520, opened 2021-08-05
Bump python Kopf S2i builder to v1.33.0

jkupferer:
#### What is this PR About?

Increase Python Kopf version to 1.33.0.

#### How do we test this?

Test as described in the build-s2i-python-kopf README using examples provided.


pabrahamsson (approved):
/lgtm

PR #522, opened 2021-08-24
🐩 cosign Jenkins agent added 🐩

ckavili:
#### What is this PR About?
Extend Jenkins agent with cosign binary.

#### How do we test this?

### DIFF
--- a/jenkins-agents/jenkins-agent-cosign/Dockerfile
+++ b/jenkins-agents/jenkins-agent-cosign/Dockerfile
@@ -1,0 +1,11 @@
+FROM quay.io/openshift/origin-jenkins-agent-base:4.8
+
+USER root
+
+ARG COSIGN_VERSION=1.0.0
+
+# Install cosign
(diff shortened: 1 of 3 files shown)


jenkins-agents/jenkins-agent-cosign/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM quay.io/openshift/origin-jenkins-agent-base:4.8
```
springdo, line 1: @ckavili  - I think the other agents are based on 4.7 so we might need to update those separately ....

springdo (approved):

springdo:
d'oh! I forgot the lgtm would merge this automatically .... 

I think we need to add an entry for this guy in here somewhere? https://github.com/redhat-cop/containers-quickstarts/blob/master/.applier/group_vars/seed-hosts.yml 

@garethahealy  will know better though. Are these still used or are we all GH ACtions now?

PR #525, opened 2021-09-16
Update Jenkins Agent base images to 4.8 + minor improvements

pabrahamsson:
#### What is this PR About?
- update Jenkins Agents to use a 4.8 base image
- remove usage of `-k` with `curl` in Jenkins Agent Dockerfiles
- use `curl <url> | tar zxf - -C <path>` pattern to reduce `Dockerfile` size
- Default to use latest stable version of `oc` client (when installed)
 
#### How do we test this?
Should be covered by CI

Can also be tested manually
```bash
./_test/setup.sh applier container-quickstarts-tests pabrahamsson/containers-quickstarts feature/jenkins-agent-4.8-base
./_test/setup.sh test container-quickstarts-tests pabrahamsson/container-quickstarts feature/jenkins-agent-4.8-base
```

### DIFF
--- a/jenkins-agents/jenkins-agent-ansible/Dockerfile
+++ b/jenkins-agents/jenkins-agent-ansible/Dockerfile
@@ -1,9 +1,9 @@
-FROM quay.io/openshift/origin-jenkins-agent-base:4.7
+FROM quay.io/openshift/origin-jenkins-agent-base:4.8
 ARG ANSIBLE_VERSION=2.9.13
 
 LABEL \
       release="1" \
-      version="4.7" \
+      version="4.8" \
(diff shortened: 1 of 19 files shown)


springdo (approved):

springdo:
@pabrahamsson - should i be waiting for `cq-jenkins-agents-pr-build` to do something before adding the label?

pabrahamsson:
> @pabrahamsson - should i be waiting for `cq-jenkins-agents-pr-build` to do something before adding the label?

It shouldn't matter when you add the label, if the pr build job fails it will not be merged (even if all required labels are added)

springdo (approved):
/approve

PR #536, opened 2021-11-11
Fix broken permissions in Jenkins agents using Python 

pabrahamsson:
#### What is this PR About?
The jenkins-agent-ansible image installs `python38` which through dependencies also updates `chkconfig`. `chkconfig` will update the permissions on `/var/lib/alternatives` and `/etc/alternatives` (`755`), this breaks the `alternatives` command (called as part of the `ENTRYPOINT`).
```
bash-4.4$ /usr/local/bin/run-jnlp-client
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
failed to create /var/lib/alternatives/java.new: Permission denied
```

This PR will "reset" the permissions on `/var/lib/alternatives` and `/etc/alternatives` to the [upstream settings](https://github.com/openshift/jenkins/blob/release-4.8/slave-base/Dockerfile.localdev#L34-L35).

It will also add a test to ensure the `java` executable is in good order.

#### How do we test this?
CI should ensure we now have a working jvm.
Can also test locally:
```
git clone https://github.com/pabrahamsson/containers-quickstarts
cd containers-quickstarts/jenkins-agents/jenkins-agent-ansible
git checkout fix-jenkins-agent-ansible-java
podman build -t jenkins-agent-ansible .
podman run -it --rm -u 1001:0 jenkins-agent-ansible:latest
```
There should be no complaints about permissions on `/var/lib/alternatives/java.new`.

### DIFF
--- a/jenkins-agents/jenkins-agent-ansible/Dockerfile
+++ b/jenkins-agents/jenkins-agent-ansible/Dockerfile
@@ -34,7 +34,9 @@ RUN set -x \
      && rm -rf /var/cache/dnf \
      && alternatives --set python3 /usr/bin/python3.8 \
      && python3 -m pip install ${PIP_PKGS} \
-    && echo
+     && chmod -R 775 /etc/alternatives \
+     && chmod -R 775 /var/lib/alternatives \
+     && echo
 
 USER 1001
 ENV \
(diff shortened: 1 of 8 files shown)


sabre1041 (approved):
LGTM

PR #541, opened 2021-11-19
Feature: ubi8-asciidoctor plantuml support

pabrahamsson:
#### What is this PR About?
Add openjdk-11 + python dependencies.
This potentially resolves #530 

The following asciidoc seems to generate a proper plantuml diagram
```
[plantuml,jdot,png]
----
!pragma layout smetana
class RedHat
class Fedora
class OpenShift
class Ansible
RedHat <|-- OpenShift
RedHat <|-- Ansible
RedHat <|-- Fedora
----
```
image

#### How do we test this?
See above

cc: @redhat-cop/day-in-the-life

PR #542, opened 2021-12-01
Update to ubi8 latest and latest pandoc

buuhsmead:
#### What is this PR About?
Update to latest ubi8 to resolve vulnerabilities.
Update to latest pandoc.
Also update to version 1.10 because of PR #541 
Which I also tested with this version.

#### How do we test this?
I take an older generated pdf and compare it with a new generated pdf with diff-pdf.


pabrahamsson (approved):
/lgtm

PR #545, opened 2022-01-18
removed tyle and patrick as they've left RH

garethahealy:
#### What is this PR About?
$subject


pabrahamsson (approved):
/lgtm

PR #549, opened 2022-02-04
Add helm binary to Tower image

jfilipcz:
#### What is this PR About?
This PR adds helm binary to Tower image template.

#### How do we test this?
Build the image out of template and verify helm is available

### DIFF
--- a/tower-ocp-custom/.openshift/templates/tower-ocp-custom-image.yml
+++ b/tower-ocp-custom/.openshift/templates/tower-ocp-custom-image.yml
@@ -27,6 +27,7 @@ objects:
                 https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
           RUN yum install -y \
                 python-devel \
+                helm \
                 pandoc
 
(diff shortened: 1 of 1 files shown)


springdo (approved):
/lgtm

oybed:
@jfilipcz thx for submitting this. However, `helm` is already part of the upstream image, so this change isn't quite needed.

```
> oc get deployment/tower  -o yaml | grep 'image: registry'
      - image: registry.redhat.io/ansible-tower-38/ansible-tower-rhel7:3.8.2
> oc debug deployment/tower
Starting pod/tower-debug ...
helm version
Pod IP: 10.128.2.29
If you don't see a command prompt, try pressing enter.
sh-4.2$ helm version
version.BuildInfo{Version:"v3.5.0+3.el7at", GitCommit:"77fb4bd2415712e8bfebe943389c404893ad53ce", GitTreeState:"clean", GoVersion:"go1.14.12"}
```

One useful functionality, however, would be to stay consistent with the other tooling we use on how we deploy these tools, and hence can specify a specific helm version to use (i.e.: an override). 

For now, I'd say we should revert this PR, but consider additional functionality in the future if needed.

PR #550, opened 2022-02-07
[Tower Custom Image] Revert addition of helm binary

jfilipcz:
This PR reverts the helm installation added for Tower image template,  applied in https://github.com/redhat-cop/containers-quickstarts/pull/549 , which was not needed and was applied by mistake

#### What is this PR About?
Removal of helm binary installation in custom Tower image template

#### How do we test this?
Run the Tower image build and confirm helm is not installed during the build.


oybed (approved):
/lgtm

PR #557, opened 2022-02-15
added ghcr to publish ci and remove branch filtering to always publish on push

garethahealy:
#### What is this PR About?
Reason/concept behind this change: when building on my fork, as the actions are only active on master, I never know if my changes have worked as I normally work on a dedicated branch for that change. this change removes those filters for that reason.

also, as we push to quay, my fork needs to have quay config setup. i have this already, but 99% of people wont. so I've added ghcr in so they can run the publish

#### How do we test this?
CI goes green on merge. you can see it green on my fork: https://github.com/garethahealy/containers-quickstarts/actions

### DIFF
--- a/.github/workflows/build-s2i-python-kopf-publish.yaml
+++ b/.github/workflows/build-s2i-python-kopf-publish.yaml
@@ -2,8 +2,6 @@
 name: build-s2i-python-kopf-publish
 on:
   push:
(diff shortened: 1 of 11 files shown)


pabrahamsson (approved):
I like this, thank you @garethahealy 
/lgtm

Issue #560, opened 2022-02-16
hygieia fails to build on ci


garethahealy:
- https://github.com/redhat-cop/containers-quickstarts/runs/5214103164?check_suite_focus=true

PR #564, opened 2022-02-16
updated base image to fix yum failure on build

garethahealy:
#### What is this PR About?
fixes: https://github.com/redhat-cop/containers-quickstarts/issues/560


pabrahamsson (approved):
/lgtm

Issue #561, opened 2022-02-16
s2i-mvn-gradle fails to build on ci


garethahealy:
- https://github.com/redhat-cop/containers-quickstarts/runs/5214102989?check_suite_focus=true

PR #565, opened 2022-02-16
updated base and mvn cdn url to fix build

garethahealy:
#### What is this PR About?
fixes: https://github.com/redhat-cop/containers-quickstarts/issues/561


pabrahamsson (approved):
/lgtm

PR #567, opened 2022-02-16
replaced version code with central version action

garethahealy:
#### What is this PR About?
$subject

#### How do we test this?
ci stays green


pabrahamsson (approved):
It somewhat rubs me the wrong way seeing the tekton tasks using `VERSION` instead of `version.json` but that's a PR for a different day.
/lgtm

PR #572, opened 2022-03-16
ubi8-asciidcotor - add asciidcotor-reducer gem

itewk:
#### What is this PR About?
add the asciidcotor-reducer gem which replaces the old asciidoctor-coaleser script.


pabrahamsson (approved):
/lgtm

PR #574, opened 2022-04-07
Add initial KinD implementation for testing jenkins agents

pabrahamsson:
#### What is this PR About?
This will add a step to the GitHub workflows for the Jenkins Agents to test them by using KinD (Kubernetes in Docker).

#### How do we test this?
This should be handled by the GH actions on this repo

### DIFF
--- a/_test/kind/setup.sh
+++ b/_test/kind/setup.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+AGENT=$1
+JENKINS_CHART_VERSION=${2:-3.11.10}
+AGENT_PATH="jenkins-agents/${AGENT}"
+SCRIPT_DIR=$(dirname -- "$(readlink -f "${BASH_SOURCE[0]}" || realpath "${BASH_SOURCE[0]}")")
+
+if [[ -d ${AGENT_PATH} ]]
+then
+  # Create KinD cluster and load required container images
+  kind create cluster --config ${SCRIPT_DIR}/kind-config.yaml
+  for image in kiwigrid/k8s-sidecar:1.15.0 jenkins/jenkins:lts-jdk11 ${AGENT}:latest
+  do
+    if [[ "${image}" =~ "${AGENT}" ]]
+    then
+      podman save ${image} | docker load
+      docker tag localhost/${image} ${image}
+    else
+      docker pull docker.io/${image}
+    fi
+    kind load docker-image ${image}
+  done
+  # Create Nginx Ingress controller
+  kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
+  echo "### Wait for Ingress controller to install ###"
+  kubectl wait --namespace ingress-nginx \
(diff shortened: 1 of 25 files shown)


_test/kind/setup.sh
```diff
@@ -0,0 +1,108 @@
...
+    let "timeout += 2"
+  done 
+
+  echo "Build in progress..."
+  timeout=0
+  while [[ $(curl -s http://localhost/job/containers-quickstarts/job/${AGENT}/lastBuild/api/json --user admin:${token} | jq -r '.building') == "true" ]]
+  do
+    if [[ ${timeout} -gt 300 ]]
+    then
+      echo "Timed out waiting for build to finish..."
```
garethahealy, line 92: would be nice if these timeout breakouts dumped some logs. happy for that to be solved in another pr.

_test/kind/setup.sh
```diff
@@ -0,0 +1,108 @@
...
+
+AGENT=$1
+JENKINS_CHART_VERSION=${2:-3.11.10}
+AGENT_PATH="jenkins-agents/${AGENT}"
+SCRIPT_DIR=$(dirname -- "$(readlink -f "${BASH_SOURCE[0]}" || realpath "${BASH_SOURCE[0]}")")
+
+if [[ -d ${AGENT_PATH} ]]
+then
+  # Create KinD cluster and load required container images
+  kind create cluster --config ${SCRIPT_DIR}/kind-config.yaml
```
garethahealy, line 13: cant decide whether we should use the kind github action to handle all the kind stuff or not. happy for it to be merged as-is, just sharing my current thinking.

_test/kind/setup.sh
```diff
@@ -0,0 +1,108 @@
...
+      echo "Timed out waiting for build to be created..."
+      exit 1
+    fi
+    sleep 2
+    let "timeout += 2"
+  done
+
+  echo "Waiting for build to start..."
+  timeout=0
+  until [[ $(curl -s http://localhost/job/containers-quickstarts/job/${AGENT}/lastBuild/api/json --user admin:${token} | jq -r '.building') == "true" ]]
```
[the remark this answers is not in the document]
pabrahamsson: yes, same check but line 75 is waiting for the build to start and line 88 is waiting for the build to finish. Hence the use of `until` and `while` respectively.
I agree that the code feels very repetitive so definitely open to any suggestions for improvement(s).

_test/kind/setup.sh
pabrahamsson: I briefly thought about this as well but given the one line used here I decided against the kind GH action for now. Definitely open to re-evaluate.

_test/kind/setup.sh
pabrahamsson: Agreed, I'll add the build logs to the build timeout but not sure what we could/should log for the job start timeout?

pabrahamsson:
@garethahealy I've responded to your comments and also made a few updates. PTAL when you have a chance.

garethahealy:
am happy to merge as-is, and look at improving with future PRs.

garethahealy (approved):

PR #586, opened 2022-09-08
WIP: Add optional Hosted Zone ID parameter to ClusterIssuer template

jfilipcz:
#### What is this PR About?
This PR add an optional Hosted Zone ID parameter to ClusterIssuer template on Cert-Manager configuration 

#### How do we test this?
Deploy ClusterIssuer on Openshift running Cert-Manager


pabrahamsson (commented):
Looks good @jfilipcz 
Should we make the same change to the Helm chart to keep them in sync?

jfilipcz:
It won't hurt to update the helm part, but I still need to do some testing on the template itself (I'm not sure it won't break the Clusterissuer in other places we're using that template)

pabrahamsson (approved):
LGTM

PR #590, opened 2022-11-17
Update cosign from v1.0.0 to v1.7.2 on jenkins-agent-cosign

jfilipcz:
#### What is this PR About?
This PR updates jenkins-agent-cosign Dockerfile to install v1.7.2 version of cosign instead of v1.0.0

#### How do we test this?
Build&Run the image with TL500 exercise


pabrahamsson (approved):
LGTM

PR #595, opened 2023-01-20
Update components using generate with ascii-doctor

buuhsmead:
#### What is this PR About?
Update of components inside: asciidoctor and Pandoc

#### How do we test this?
Tested local with local build via podman that the generated pdf follows expected layout, which this version does.


garethahealy:
@buuhsmead ; can you rebase to get rid of the old commits? can see stuff from 2021 which is a bit confusing when looking at the PR

buuhsmead:
Would love to do that, but git skills are not up to par. tried 'git rebase -i <sha-after-Merge branch 'redhat-cop:master' into master> but that doesn not bring me into the right state. Sorry

buuhsmead:
Will start over. Takes too much time for you.

PR #604, opened 2023-10-11
Fedora based image for Asciidoctor 

buuhsmead:
#### What is this PR About?
I would like to add a Fedora based asciidoctor utility image besides the already available ubi8-asciidoctor.
It is much easier to keep this image up2date, as the version dependency comes complete from Fedora:latest.
For updating the image, only the 'version.json' needs a newer version number

#### How do we test this?
Create the image.
Generate a pdf.
Compare with previous generated pdf, from previous image version of ubi8-asciidoctor.
Tested this version with older generated pdf.
pdf is generated and comparable with previous version


buuhsmead:
I made a mistake. Too much old commits. Sorry.

PR #630, opened 2023-12-11
marked old content as deprecated

garethahealy:
@pabrahamsson 


springdo:
I have a funny feeling that the `ipa-server` one is used by something still Gareth, perhaps used for the Demo platform or for the old labs automation.... @eformat  / @ckavili  / @oybed might also know as well as Petter 🤷

garethahealy:
@springdo reverted ipa move

pabrahamsson (dismissed):
LGTM

garethahealy:
reverted move and added to readme

OWNERS
```diff
@@ -1,20 +0,0 @@
-# See the OWNERS docs at https://go.k8s.io/owners
```
oybed, line 1: Can we leave this file in place? ... alternatively update it if there are users that are no longer involved?

garethahealy: as we don't use prow anymore, this file doesn't do anything thats why I've replaced it with `CODEOWNERS` which uses the github configured teams.

oybed: Understood, but we also look at this file in a manual way, so by removing it we loose some of the valuable info. Is there a way to add a "default" to the `CODEOWNERS` file to keep the rest of the users from this file?

oybed: Cool - missed the "default" / * at the top of the `CODEOWNERS` file. There's a discrepancy between that group and the old file, but we can work on that separately.

PR #631, opened 2023-12-12
updated tekton versions to be renovated

garethahealy:
- added renovate to tekton images
- added default permissions for CI to be read-only, and elevated where needed - https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions


pabrahamsson (approved):
LGTM

PR #633, opened 2023-12-12
added default perms for ci on utilities

garethahealy:
$subject - https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions


pabrahamsson (approved):
LGTM

PR #642, opened 2023-12-12
added exclude for push when its renovate

garethahealy:
think this should fix renovate pushing on PRs


pabrahamsson (approved):
LGTM

PR #683, opened 2023-12-18
cleaned up jenkins agents to use common base image

garethahealy:
- added hadolint to lint dockerfiles
- added multi stage builds, so any tars are pre-downloaded and only required bits are copied into runnable image
- updated to use common base agent


garethahealy:
good to merge

pabrahamsson (approved):
LGTM, thank you for doing this

PR #794, opened 2024-05-16
Updated versions of tools for tool-box

oybed:
#### What is this PR About?
Updated versions of tools in the `tool-box` container

### DIFF
--- a/tool-box/Dockerfile
+++ b/tool-box/Dockerfile
@@ -4,19 +4,19 @@ FROM registry.access.redhat.com/ubi9/ubi:9.3-1610@sha256:66233eebd72bb5baa25190d
 SHELL ["/bin/bash", "-c"]
 
 # renovate: datasource=repology depName=homebrew/openshift-cli
-ARG OC_VERSION=4.14.8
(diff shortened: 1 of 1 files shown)


tool-box/Dockerfile
```diff
@@ -4,27 +4,27 @@ FROM registry.access.redhat.com/ubi9/ubi:9.3-1610@sha256:66233eebd72bb5baa25190d
 SHELL ["/bin/bash", "-c"]
 
 # renovate: datasource=repology depName=homebrew/openshift-cli
-ARG OC_VERSION=4.15.11
+ARG OC_VERSION=stable-4.15
 # renovate: datasource=github-releases depName=openshift/rosa
-ARG ROSA_VERSION=v1.2.37
+ARG ROSA_VERSION=latest
```
garethahealy, line 9: would rather keep these pinned to versions, that's the whole point of letting renovate update

oybed: sure - but the bigger problem is that there's very infrequent releases of this repo

PR #930, opened 2025-04-21
Python kopf s2i v1.37.5

jkupferer:
#### What is this PR About?

- Update kopf to v1.37.5
- Update other components and dependendencies to latest tags

#### How do we test this?

Tested with examples in repository.


garethahealy:
@jkupferer CI is failing, still want this merging?

This is only my view, but maybe a good idea to start thinking about a new home for this. This repo is inactive and probably will be archived at some point in the future.

garethahealy:
stale

PR #968, opened 2025-05-13
Update package versions

branic:
#### What is this PR About?
Updates versions for the ubi8-asciidoctor image

#### How do we test this?
Build the image using the Dockerfile and generate a PDF from an asciidoc source using asciidoctor-pdf command.


garethahealy (approved):
/lgtm

